PR #9, opened 2026-09-12
Add file lock and recovery to fix EOFError on concurrent requests

### LINKED ISSUE #4: Fix issues with DB opening
The seriousdb currently does not work (see bug faced at the end of the video https://www.youtube.com/watch?v=VkeDvjHbubg), because pickle is encountering a (for me) unexpected EOF.

UniquePython:
Fixes #4

The PUT and GET handlers open and close the .sdb file independently with no synchronization. Under concurrent requests, one request's write can truncate the file while another request is mid-read, which throws EOFError on pickle.load. Wrapped the read-modify-write cycle in a lock so requests are serialized instead of racing.

Also added a load_db() helper that rebuilds the file with default values if it's ever found missing, empty, or corrupted, so an already-corrupted .sdb from a previous run recovers on the next request instead of crashing every time.

Removed the "database file could not be opened" branch in GET since it could never actually run - pickle.load throws before that check is reached.


danieldeer:
I have replaced pickle with JSON. I think the JSON module does not have the EOF error. Can you double check?

UniquePython:
> I have replaced pickle with JSON. I think the JSON module does not have the EOF error. Can you double check?

Just tested it against the current main.py (the json + readline version) - switching to json doesn't fix this, it just changes which exception shows up. The race is in the read-modify-write pattern itself, not the serialization format: if two requests overlap, one's write can truncate the file while another's read is mid-file. With pickle that's `EOFError`; with the current json code it's `json.decoder.JSONDecodeError` (`"Expecting value: line 1 column 1"`).

Reproduced it using the exact logic currently in main.py's put():

```python
def put(key, value):
        binary_text = f.readline()
        json_text = binary_text.decode()
        db = json.loads(json_text)
        db[key] = value
    with open(db_file, "wb+") as f:
        json_dumps = json.dumps(db).encode()
        f.write(json_dumps)
```

Ran two of these concurrently in threads and got the JSONDecodeError above within a few runs.

The lock in this PR fixes it regardless of serialization format, since the actual bug is unsynchronized file access, not pickle vs json. Happy to update the diff to apply the lock on top of the current json-based code, or if #26's cache-based rewrite already covers this, I'm fine closing mine in favor of that - whichever direction you'd rather go.

danieldeer:
Since #26 provides caching, which we definitely need, I am going to prefer #26! Thanks for your analysis.
Can you confirm that #26 also fixes the issue through its locking mechanism?

UniquePython:
Confirmed, tested #26 and it resolves the concurrency issue. One note: it's still on pickle rather than json, so merging would revert #11's fix.

danieldeer:
Fixed already. Thanks! Closing this.

PR #15, opened 2026-09-12
feat(auth): Adding a very simple template for a future auth system

Airbag65:
### Authentication

It might be a good idea to not let just anyone hit the endpoints as they please. This adds a very simple working template for how a future, more robust authentication system might look.

#### Usage
- If you don't want you use any authentication at all, just toggle the `require_auth` flag in `auth.py` to false. If you do want authentication, toggle it to true.
- You'll need to set an `auth_token` in a `.env` (this PR comes with an example of that)
- In the request, use the `Autorization` header with the typical `Bearer <token>` pattern

If the token is not correct or the usage is not correct, you'll receive a `401 Unauthorized` response. Otherwise the endpoints will respond with what they usually do.


main.py
```diff
@@ -13,9 +15,12 @@
 
 app = FastAPI()
 
+
 @app.put("/db")
-async def put(key: str, value: str):
+async def put(request: Request, key: str, value: str):
     db = None
+    if not is_validated(request):
```
crydafan: How about using an authentication middleware instead? For a cleaner separation of concerns.

Here's a simple example from Gemini:

```python
import time
from fastapi import FastAPI, Request

app = FastAPI()

@app.middleware("http")
async def add_process_time_header(request: Request, call_next):
    # 1. Code here runs BEFORE the request hits the route
    start_time = time.time()
    
    # 2. Pass the request down the line to the endpoint
    response = await call_next(request)
    
    # 3. Code here runs AFTER the route finishes (on the response)
    process_time = time.time() - start_time
    response.headers["X-Process-Time"] = str(process_time)
    
    return response

@app.get("/")
async def root():
    return {"message": "Hello World"}
```

mattsva: +1

config.json
```diff
@@ -0,0 +1,3 @@
+{
```
mattsva: @danieldeer is it even supposed to use json as a config lang.? Maybe we could use python itself (would make sense) or something else?
If you say json is good, we could stick at it, I mean you are the maintainer.

auth.py
```diff
@@ -0,0 +1,32 @@
+import json
+import os
+
+from fastapi import Request
+from dotenv import load_dotenv
+
+
+def load_config():
```
mattsva: what if there is no config file?
-> I would make a fallback here.

mattsva: I mean, yeah, if there’s no configuration, this may be unnecessary. But if we want to keep this config file for some important settings, then I think this should be necessary.

.gitignore
```diff
@@ -98,7 +98,7 @@ ipython_config.py
 #   Similar to Pipfile.lock, it is generally recommended to include uv.lock in version control.
 #   This is especially recommended for binary packages to ensure reproducibility, and is more
 #   commonly ignored for libraries.
-# uv.lock
+uv.lock
```
mattsva: I dont think that this should be in this PR. Its a general problem.

main.py
```diff
@@ -25,14 +30,20 @@ async def put(key: str, value: str):
...
+async def get(request: Request, key: str):
     db = None
+    if not is_validated(request):
+        raise HTTPException(status_code=401, detail=f"Not authenticated")
     with open(db_file, "rb") as f:
         db = pickle.load(f)
     f.close()
     if db is None:
-        raise HTTPException(status_code=404, detail=f"Database file {db_file} could not be opened and loaded")
+        raise HTTPException(
```
mattsva, line 55: @Airbag65 why change this?

main.py
Airbag65: Yes, I like that idea. Will amend the PR with this when I get the chance to

mattsva (changes requested):
I though about maybe even making a authentication so that it can be turned on / off per separate data, and not just for all. But this is just an idea, need a approve from @danieldeer.

I said what I would change in the code comments.

main.py
mattsva: @Airbag65 You can change the PR on fly, with another commit and pull to your fork, in the right branch, so you could do that here if you like to.

main.py
crydafan: Possibly the formatter

main.py
Airbag65: @mattsva That is what I had in mind

config.json
Airbag65: Not much thought actually went into the decision to choose JSON. I did think briefly to use TOML, but for a working POC i just went with what was the easiest for me to make work in the moment

main.py
mattsva: I tested it now, because I was curious. Yeah black does that.

auth.py
Airbag65: A fallback is absolutely necessary in this case, I'll add that! A config-file for for just this case might very well be overkill, but I thought it might be a good thing to have for the future in order to actually have a place to add more "settings"

.gitignore
Airbag65: Agreed!

auth.py
mattsva: Im agreeing on that. Its just necessary to choose the right lang to configure. In this simple case it doesnt matter that much but maybe is over all a lang like python better to configure. So as example a `configuration.py` except for a json. But thats something @danieldeer needs to decide mainly.

config.json
mattsva: Sure, json isnt wrong. I would personaly just use Python for that, or maybe Lua, but Lua would be not good, because we use Python in the project itself, that would be convenient. But another lang would be fine too. It depence on the complexity those settings will get over time.

mattsva: With a python `configuration.py` you could make filtering smart. So as example if an obj id starts with 'admin.' its just for API keys with the admin permission, if its starting with 'base.' its just the baseline permission.
I hope everyone gets what I mean now xd

Airbag65: I absolutely see your point, and agree. Will shortly be adding another commit to the PR where this will be addressed :)

configuration.py
```diff
@@ -0,0 +1,13 @@
+from dataclasses import dataclass
```
mattsva: Thats good for structuring the code, but I would even consider having a dedicated configuration language, using Python as an example. This is just a blueprint, but it could look something like this:

```python
from seriousdb import config

@config
class AUTH:
    if id.startswith("admin."):
        require(permission.admin)
    elif id.startswith("user."):
        require(permission.user)
    else:
        require(None)
```

Don't get me wrong, this isn't necessarily a perfect approach, but I hope it makes the idea clear.

A `.env` file is useful, but I don't think it's the right abstraction for everything. We could support configuration through `.json`, `.py`, `.toml`, or even `.env`, depending on the use case. However, I would strongly disagree with making `.env` the primary solution for more complex configuration.

Personally, I would prefer a `.py`-based configuration language, or some kind of configuration language integrated into the DB itself - potentially even supporting both. A real programming language makes the possible use cases and usability much wider than a static format like `.env` or `.json`. Static configuration isn't always the best solution; having fluid/configurable logic becomes increasingly useful as the project grows and gains more features.

Python's data structures and overall data-management capabilities are especially interesting for this, particularly if the application itself is written in Python like we have it here.

Nevertheless, an interface like the current `configuration.py` could still be useful. I would see it more as an interpreter/interface for the actual configuration file, rather than necessarily being the configuration format itself.

The main point is that I don't think this is really about auth anymore. It's a more fundamental decision about what should be configurable in the final project and how configuration should work.

Because of that, I wouldn't try to solve this in the current PR. I would rather open a separate PR specifically for a configuration language. I'm considering working on that, but I'm also completely open to you doing it. You may have a different approach in mind, or I may be misunderstanding something, so please correct me if that's the case.

I'd also like to discuss this with @danieldeer too.

Should we open a discussion specifically about the configuration system/language? I think it would be useful to settle the general direction before implementing it.

crydafan: The problem with a dynamic configuration like you mentioned is that it needs to be evaluated for each request, possibly thus introducing the overhead of running the configuration logic. 

Your example reminds me of authorization, one of the standards to handle authorization is by using JWTs, these are stateless tokens that contain signed metadata issued by a user, created by the server and then sent back to the user for future request authorization.

I wouldn't mind using a static `.conf` based configuration. Possibly in the future there could be a use case for dynamic configuration.

mattsva: I do think booth do have their needs. It depence on the use case and on the features. Now it's not necessary but later it may be.

Maybe the auth wasn't the best example for it, I just wanted to show that this could be smth like that.
It does add a little overhead, but as a cost you may wanna pay.
Maybe make such a think like dynamic configs optional?

mattsva: I know what JWTs are @crydafan btw

Airbag65: I actually didn't think of the fact that the config aspect of this PR probably shouldn't be included in the same PR, but I fully agree that it should probably be separate. And I do think we should open up a discussion thread for it before actually deciding, and of course @danieldeer has the last say in the question! I'll make some modification to this PR, to just simply have it as a global variable for now, and then I'll open that thread

Airbag65:
I have solved the above mentioned and opened a [discussion thread](https://github.com/danieldeer/seriousdb/discussions/41) for the configuration

receptaha:
Isn't the goal of this project to create an HTTP database? That implies each client needs its own separate database, so I assume each auth token should have its own unique database. However, I see in this PR that the token is being implemented via an environment variable; is that correct?

Airbag65:
This PR is by absolutely no means a fully fletched auth implementation. There would absolutely need to be additional things to make it work for every use case, but it has a working POC and something to work from.

I'd say the project ist not yet at a state where it would justify a complete implementation, but I welcome all feedback

configuration.py
mattsva: > I actually didn't think of the fact that the config aspect of this PR probably shouldn't be included in the same PR, but I fully agree that it should probably be separate. And I do think we should open up a discussion thread for it before actually deciding, and of course @danieldeer has the last say in the question! I'll make some modification to this PR, to just simply have it as a global variable for now, and then I'll open that thread

Good for now

Airbag65:
Made a new version of this PR to follow the new code style. Closing this one

PR #17, opened 2026-09-12
Add Nix development environment

mattsva:
# Adds a reproducible Nix development environment with Python 3.14, FastAPI, FastAPI CLI, and Black.

Use it with:
```bash
nix develop
```


mattsva:
@danieldeer  you may not need Nix flakes, but they make life for NixOS users (and Nix users in general) a lot easier. They provide the packages in a shell, so it’s kind of like uv, but even easier.

danieldeer:
@mattsva I have never seen Nix flakes in repositories, and I am sure they make people's lives easier. I just don't know whether they should be part of this repository, because they are useful only to a minority of users.
I need some expert opinions here.

crydafan:
Hey, @danieldeer

Every serious open source project needs a Nix flake. In all seriousness, this reminds me of when I was working at a startup and one of the projects had a Nix flake in it, cool.

From Gemini:

> Nix flakes are a standardized, self-contained format for managing Nix projects and configurations using a `flake.nix` file and a lock file (`flake.nix` and `flake.lock`).

Now, I'm no expert but I would believe this is quite similar to how NPM works by using a lock file to pin package versions. That's my two cents.

Cheers!

mattsva:
@crydafan Yeah, flake.lock files do work like that: they pin dependencies to an exact commit (with a hash), effectively locking the version of the package.

And @danieldeer, I don't think it's a minor part. Big projects like zed-industries/zed and hyprwm/Hyprland use them too. There are plenty of other examples, but those are the two that come to mind.

crydafan (approved):
Looks good to me!

tymscar:
More and more projects go down this path and it makes a lot of sense as using Nix is not just something people on NixOS do, but also those on any other linux distro, or macos, if they so choose.

mattsva:
> More and more projects go down this path and it makes a lot of sense as using Nix is not just something people on NixOS do, but also those on any other linux distro, or macos, if they so choose.

I meant that actually. I may use NixOS but I consider every one using Linux or maybe even Macos as a programmer using Nix and Nix flakes.
Isn't there even a BSD Port, I guess so.

PR #23, opened 2026-09-12
Improve API endpoints and validation

FriedAxons:
## Summary

This PR adds API functionality for deleting keys and retrieving all stored key-value pairs, while aligning the implementation with the project's current cache-based architecture.

### Changes

- Add `DELETE /db` to remove a key-value pair and return its value.
- Add `GET /db/all` to retrieve all currently stored key-value pairs.
- Validate that `PUT /db` keys contain at least one character.
- Integrate the API additions with the shared `Cache` implementation.
- Keep API routes in `main.py` and use the existing cache methods for database operations.
- Preserve the project's JSON persistence, locking, lifespan handling, and background flush behavior.
- Add explicit cache type annotations to address static type-checking issues.

### API behavior

`DELETE /db` returns the deleted value and a `404` response when the key does not exist.

`GET /db/all` returns all currently stored key-value pairs.

`PUT /db` returns `422` when the key is empty.


mattsva (approved):
lgtm

danieldeer:
I think the DELETE endpoint is needed.
We don't need a GET ALL (at least I can not think of a use case for it). In which case would you ever query a WHOLE database?
And I also would discard the empty string special handling (key validation), because in my opinion users should be able to specify an empty key. We don't know their use case. Possibly someone will create a database of hashes, and he wants to also store the hash of an empty string.

danieldeer:
Please go ahead with implementing the DELETE. Note that this needs to be aligned with the latest changes (replacement of pickle with json, and use the in-memory cache).

mattsva:
> I think the DELETE endpoint is needed. We don't need a GET ALL (at least I can not think of a use case for it). In which case would you ever query a WHOLE database? And I also would discard the empty string special handling (key validation), because in my opinion users should be able to specify an empty key. We don't know their use case. Possibly someone will create a database of hashes, and he wants to also store the hash of an empty string.

I think an GET ALL could be useful. As you said "We don't know their use case.", so maybe leave that in for now.

danieldeer:
okay, @mattsva convinced me

FriedAxons:
> I think the DELETE endpoint is needed. We don't need a GET ALL (at least I can not think of a use case for it). In which case would you ever query a WHOLE database? And I also would discard the empty string special handling (key validation), because in my opinion users should be able to specify an empty key. We don't know their use case. Possibly someone will create a database of hashes, and he wants to also store the hash of an empty string.

I’ve updated the API changes to align with the JSON persistence and in-memory `Cache` approach you introduced.

The `GET /db/all` and `DELETE /db` endpoints now use the shared cache and locking implementation rather than the previous direct pickle file access. I also cleaned up the cache state with explicit type annotations and handled the optional filename/database values so the type-checking warnings around `open()` and `cache.db` are resolved.

Regarding the API design feedback, I agree that `DELETE` is useful. I can also see your point about allowing an empty string as a valid key, since we shouldn’t assume what users will use the database for. I’m open to removing that validation.

For `GET /db/all`, I added it mainly because it is useful for inspecting the contents of this small key-value store during development and testing, but I understand the concern about exposing a whole-database query. I'm happy to remove it if you think keeping the API surface minimal is preferable.

Thanks for the feedback on both the persistence direction and the API design. It helped me bring the changes in line with the project's newer approach rather than continuing with the older pickle-based implementation.

mattsva (approved):
LGTM. I did not test it, but it looks right.

mattsva:
Does this conflict with newer PRs like https://github.com/danieldeer/seriousdb/pull/63 or other restructuring? 

Nevertheless, it looks good

FriedAxons:
> Does this conflict with newer PRs like #63 or other restructuring?
> 
> Nevertheless, it looks good

It has been updated to reflect those changes. I am now working on updating the test suite to also match those changes for #27 and resolve conflicts.

FriedAxons:
@danieldeer I've updated PR #23 to align the API changes with the current modular JSON/cache architecture.

The `GET /db/all` and `DELETE /db` endpoints now use the shared `Cache` and locking implementation, with database operations kept in `db.py` and HTTP routes in `main.py`. I also updated the cache typing so the optional filename and database state are explicit and the related Pylance warnings are resolved.

I also kept the existing key validation for now. I saw your point about allowing an empty string as a valid key, and I'm happy to remove that behavior if you prefer empty keys to be supported.

For `GET /db/all`, I understand the concern about exposing a whole-database query. I added it primarily because it is useful for inspecting this small key-value store during development and testing, but I'm happy to remove it if you would prefer a smaller API surface.

Thanks again for the feedback. The changes now follow the project's current modular JSON/cache direction rather than the older pickle-based implementation.

src/seriousdb/db.py
acai10, line 1: Could we centralize the cache.db is None check to avoid repeating the same error handling in every function, something like get_db(cache: Cache)? Also, for select and delete_key, using direct dictionary access or pop() with a targeted except KeyError would make the distinction between a missing key and a value of None clearer.

FriedAxons: Hopefully this is more in line with what you meant. Please let me know if there is any issues. I'm learning as I go.

acai10: lgtm

FriedAxons: Thank you for your insight. In what little bit I've used Python with this project I've already learned a lot. Things in which I've never done with it before. It's also normally not the language I've primarily worked with so it's been interesting in that respect too.

acai10: I just found out about this project through Daniel's video, but I have no idea about the conventions of the codebase. Maybe we should start commenting out the code or that we add a comment to each function using docstrings, such as: 
```
    “”"
    Remove a key from the cache and return its previous value.

    Raises:
        HTTPException: If the database is unavailable or the key does not exist.
    “”"
```
Or we could leave it as is and just say that good code explains itself, but that’s up to @danieldeer to decide.

FriedAxons:
@danieldeer I’ve updated PR #23 again to align the API changes with the latest upstream architecture.

The API additions now use the current `Cache` implementation directly, including `GET /db/all` and `DELETE /db`, while preserving the key validation. I also adapted the routes to the newer lifespan, dependency-injection, and background flush patterns from upstream.

I removed the old `db.py` approach because the database operations have since been moved into `Cache`, and kept the JSON persistence, locking, and corrupt-file handling from the current implementation.

I also kept the cache type annotations that were needed to resolve the Pylance issues around `db` and `filename`.

The PR should now be aligned with the latest structure on `main`.

src/seriousdb/cache.py
AnnoDomine, line 1: @FriedAxons This is a great improvement, really liking the proper typing!

Could you also add explicit return type hints to all function/method definitions where applicable? That would make the type annotations fully complete and accurate.

AnnoDomine (changes requested):

src/seriousdb/cache.py
FriedAxons: Done. I added explicit return type hints to the applicable API and cache functions, including the lifespan handler and cache `load`/`flush` methods.

I also rebased the branch onto the latest `main` so these changes are now aligned with the current project architecture.

AnnoDomine (approved):

PR #25, opened 2026-09-12
added a simple cache class for cleaner management

Lazy690:
This PR introduces a Cache class to reduce redundant file I/O by keeping the database in memory during server operation. The implementation includes utility methods (insert, select, flush, load) to streamline common operations.

I have experience building SQL-like database engines from scratch in C++, and I'd be happy to refine this further or take on additional improvements if it aligns with the project's direction. There's no pressure to merge—I'm here to help however works best for you.


main.py
```diff
@@ -3,37 +3,55 @@
...
-    db = None
-    with open(db_file, "rb") as f:
-        db = pickle.load(f)
-        db[key] = value
-    f.close()
-    with open(db_file, "wb+") as f:
-        pickle.dump(db, f)
-    f.close()
+    insert(key, value, cache)
+    flush(cache)
```
crydafan, line 51: I may be overengineering this,

But could calling `flush` after multiple write operations become a performance bottleneck? It would make sense to enqueue flush requests and let a background worker handle them asynchronously, so writes are not blocked waiting for the flush to complete.

What happens if you overwrite an existing key while its previous value is still being flushed to disk? For example, suppose the previous value is larger than the new value and therefore takes longer to flush. The in-memory value could already have been replaced while the write of the old value is still in progress. Possibly introducing a race condition.

PR #26, opened 2026-09-12
Refactor/cache class *fixed potential race condition

Lazy690:
This PR introduces a Cache class to reduce redundant file I/O by keeping the database in memory during server operation. The implementation includes utility methods (insert, select, flush, load) to streamline common operations.

I have experience building SQL-like database engines from scratch in C++, and I'd be happy to refine this further or take on additional improvements if it aligns with the project's direction. There's no pressure to merge—I'm here to help however works best for you.


main.py
```diff
@@ -2,38 +2,63 @@
...
-    db = None
-    with open(db_file, "rb") as f:
-        db = pickle.load(f)
-        db[key] = value
-    f.close()
-    with open(db_file, "wb+") as f:
-        pickle.dump(db, f)
-    f.close()
+    insert(key, value, cache)
+    flush(cache)
```
crydafan, line 67: I may be overengineering this,

Calling `flush` after multiple write operations may become a performance bottleneck. It would make sense to enqueue flush requests and let a background worker handle them asynchronously, so writes are not blocked waiting for the flush to complete.

And also, you spotted that possible race condition from your previous PR. Good!

Lazy690:
Thank you for the feedback sir. I will look into it tomorrow.

crydafan (approved):
Looks good to me. The flush queue could be implemented in a separate PR. Nicely done!

FriedAxons:
I actually added a concurrency regression test for the same issue in my PR #27.

The test reproduces the `EOFError` from #4 with overlapping API requests and verifies that concurrent database writes do not lose data or leave the `.sdb` file corrupted. The suite currently has 14 passing tests covering the API behavior and concurrency.

I saw that #26 is also addressing the race condition, but with the additional Cache/in-memory approach. The tests in #27 are intended to validate the observable API behavior regardless of which implementation is ultimately used, so they should still be useful for testing that approach.

I figured it would be helpful to have the regression coverage in place alongside the database changes.

crydafan:
@FriedAxons I do too think that those regression tests are useful. We should also take them into consideration.

mattsva (approved):
Same opinion as @crydafan. This should may be two PRs but for now it's fine. @danieldeer please take this PR in consideration. Looks good to me, could be a good start for a better interfacing.

danieldeer:
@Lazy690 can you update your PR to use JSON (see latest changes on main branch)? Then I'll merge it!

danieldeer:
@Lazy690 can you update your PR to use JSON (see latest changes on main branch)? Then I'll merge it!

.sdb
```diff
@@ -0,0 +1 @@
+{"default": "default", "dog": "1000"}
```
danieldeer: Don't commit the .sdb file please!

danieldeer:
There's still a merge conflict. Can't merge before this is resolved.

Lazy690:
@danieldeer I excluded .sdb from the commits, and I added the functions and class on top of your main code as to not create conflict, so that you can add them as you would like to the main code, hope this is to your liking.

danieldeer:
@Lazy690 I can't merge this as this is effectively dead code now. Please integrate your provided methods into the GET and PUT calls. I think this was done before and removed now (?)

danieldeer:
@Lazy690 if you tell me this is ready, I'll merge.

Lazy690:
@danieldeer its ready

PR #31, opened 2026-09-12
Delete

guenzel50-hue:
Addet delete


main.py
```diff
@@ -25,15 +26,41 @@ async def put(key: str, value: str):
...
+            detail=f"Database file {db_file} could not be opened and loaded",
+        )
     val = db.get(key, None)
     if val is None:
         raise HTTPException(status_code=404, detail=f"No value set for key {key}")
     return val
+
+
+@app.delete("/db")
+async def delete(key: str):
```
mattsva, line 48: This does look good and works I guess.

Its pretty similar to one feature in [#23](https://github.com/danieldeer/seriousdb/pull/23) and [#34](https://github.com/danieldeer/seriousdb/pull/34)

.sdb
mattsva, line 1: Don't commit this. And don't do this in a PR. This is the database file.

pyproject.toml
```diff
@@ -7,4 +7,5 @@ requires-python = ">=3.11"
 dependencies = [
     "black>=26.5.1",
     "fastapi[standard]>=0.141.1",
+    "main-py",
```
mattsva, line 10: I don't think this is necessary.

main.py
```diff
@@ -25,15 +26,41 @@ async def put(key: str, value: str):
...
+
 @app.get("/db")
 async def get(key: str):
     db = None
     with open(db_file, "rb") as f:
         db = pickle.load(f)
     f.close()
     if db is None:
-        raise HTTPException(status_code=404, detail=f"Database file {db_file} could not be opened and loaded")
+        raise HTTPException(
```
mattsva, line 37: This is still from black. So formating only.
It probably is changed tomorrow by @danieldeer in the new YT video.
This is the wrong place for that, not in this PR.

uv.lock
mattsva, line 1: Why?

mattsva (commented):
Maybe considering to go into [#23](https://github.com/danieldeer/seriousdb/pull/23) and comment your changes there.
That will make one really good PR, instead of 2 to 3 smaller, not as good ones!

It doesnt look bad tho.

(Yeas Ive wrote something really similar in [#34](https://github.com/danieldeer/seriousdb/pull/34))

.sdb
danieldeer: I don't get how this happens. We have .gitignored the .sdb so people don't add it.

danieldeer:
Can't consider this as it is full of unwanted files. (.sdb etc.)

.sdb
mattsva: > I don't get how this happens. We have .gitignored the .sdb so people don't add it.

Shouldnt happen anymore, this was from an early state of the `.gitignore` file.

guenzel50-hue:
Jea so im new to git and just typen git add .

Daniel Hirsch ***@***.***> schrieb am So. 13. Sept. 2026 um
12:24:

PR #32, opened 2026-09-12
Load db once at startup and impl a lock

NasriAnis:
Improves performance and safety of the file-backed database:

- Load the pickle database once at startup and keep it in memory instead of reading it from disk on every single request.
- Add a `threading.Lock` around access to the in-memory db scoping it to just the read/write critical sections (in-memory reads under lock, and the disk write on `PUT`).
- `GET` no longer touches disk at all and reads straight from memory.
- `PUT` still persists changes to disk immediately after updating the in-memory copy so the file stays in sync.


src/db.py
```diff
@@ -1,39 +1,53 @@
...
+
+def save_db(db):
     with open(db_file, "wb") as f:
-        pickle.dump({"default": "default"}, f)
-    f.close()
+        pickle.dump(db, f)
+
+
+# Load the database once into memory at startup
+db = load_db()
```
mattsva: That is good, my question (and yeah hat is even at the current version a problem), what if the DB is too large for memory? Does pickel handle that automatically?

src/db.py
```diff
@@ -1,39 +1,53 @@
...
 
 db_file = ".sdb"
+db_lock = threading.Lock()
+
+DEFAULT_DB = {"default": "default"}
+
 
-# Check if the database file exists, if not populate it
-if not os.path.isfile(db_file):
+def load_db():
```
mattsva, line 11: Looks about good

src/db.py
```diff
@@ -1,39 +1,53 @@
-import pickle
 import os
+import pickle
+import threading
+
 from fastapi import FastAPI
 from fastapi import HTTPException
 
 db_file = ".sdb"
+db_lock = threading.Lock()
```
mattsva, line 6: not bad,  @danieldeer needs to decide if he likes that or not.

main.py
```diff
@@ -1,39 +1,53 @@
...
 @app.get("/db")
 async def get(key: str):
-    db = None
-    with open(db_file, "rb") as f:
-        db = pickle.load(f)
-    f.close()
-    if db is None:
-        raise HTTPException(status_code=404, detail=f"Database file {db_file} could not be opened and loaded")
-    val = db.get(key, None)
+    with db_lock:
```
mattsva: This infant is necessary. good job

mattsva (commented):
Looks good, small but important stuff.

I would recommend changing the overall structure a bit. Making some API.py files some DB.py logic files etc. so splitting that.
-> Ig @danieldeer thats something you want to talk about more in a YT video, so how to structure.

NasriAnis:
Pushed a follow-up commit (c3a0374) moving every source file to `src/` and splitting things into db.py (data layer: load/save/lock/in-memory db), api.py (FastAPI app + routes), and main.py (uvicorn entrypoint).This addresses the structure feedback from @mattsva. Also updated docs/architecture.md to reflect the new layout.

PR #34, opened 2026-09-12
Added a DELETE endpoint

xqill275:
Added a simple delete endpoint. It takes a key as input and removes it from the DB. It then returns the value associated with that key.


mattsva:
Its pretty similar to one feature in [#23](https://github.com/danieldeer/seriousdb/pull/23). It's still does have a some good approaches, as example the docs.

Maybe considering to go into [#23](https://github.com/danieldeer/seriousdb/pull/23) and comment your changes there.
That will make one really good PR, instead of 2 smaller, not as good ones!

It does look good tho.

PR #37, opened 2026-09-13
Added HEAD endpoint

Gu11e:
Added a very simple HEAD endpoint which checks whether the provided key exists in the database (basically it's a GET but only returns the http header).


main.py
```diff
@@ -25,14 +26,35 @@ async def put(key: str, value: str):
...
+            status_code=404,
+            detail=f"Database file {db_file} could not be opened and loaded",
+        )
+    val = db.get(key, None)
+    if val is None:
+        raise HTTPException(status_code=404, detail=f"No value set for key {key}")
+    return val
+
+
+@app.head("/db")
```
mattsva: I think this is a good idea just to see if there is a entry or not. Saves resources.

main.py
```diff
@@ -25,14 +26,35 @@ async def put(key: str, value: str):
...
+async def head(key: str):
+    db = None
+    with open(db_file, "rb") as f:
+        db = pickle.load(f)
```
mattsva: Isn't there some kind of `seek()` function that would be more efficient here and allow us to access the relevant part without loading the whole thing into memory?

main.py
```diff
@@ -25,14 +26,35 @@ async def put(key: str, value: str):
...
     with open(db_file, "rb") as f:
         db = pickle.load(f)
     f.close()
     if db is None:
-        raise HTTPException(status_code=404, detail=f"Database file {db_file} could not be opened and loaded")
+        raise HTTPException(
+            status_code=404,
+            detail=f"Database file {db_file} could not be opened and loaded",
+        )
+    val = db.get(key, None)
```
mattsva: Thats formatting (automatically made by black). Its not for this PR.

docs/api.md
```diff
@@ -45,3 +45,17 @@ Alice
 ```
 
 If the requested key does not exist, the API returns a `404` response.
+
```
mattsva, line 48: Great 👍 .

docs/architecture.md
```diff
@@ -12,5 +12,5 @@ The service starts with a default entry when `.sdb` does not exist. There is no
 
 1. FastAPI receives a request.
 2. The route loads the dictionary from `.sdb`.
-3. A `PUT` updates and rewrites the file; a `GET` reads the requested value.
+3. A `PUT` updates and rewrites the file; a `GET` reads the requested value; a `HEAD` only returns the header.
```
mattsva, line 15: Great.

mattsva (changes requested):
Is there a memory efficient way to do so? maybe `seek()`

main.py
```diff
@@ -25,14 +26,35 @@ async def put(key: str, value: str):
...
+    db = None
+    with open(db_file, "rb") as f:
+        db = pickle.load(f)
+    f.close()
+    if db is None:
+        raise HTTPException(
+            status_code=404,
+            detail=f"Database file {db_file} could not be opened and loaded",
+        )
     val = db.get(key, None)
```
crydafan: Instead of retrieving the actual content from the dictionary. An idiomatic version to check the existence of a key in a dictionary is by using the `in` keyword. This could be replaced with the following snippet:

```python
if key not in db:
    raise HTTPException(status_code=404, detail=f"No value set for key {key}")
```

crydafan (changes requested):

main.py
crydafan: The changes in https://github.com/danieldeer/seriousdb/pull/26 closely resemble an actual in-memory KV database mechanism. The database file is loaded at startup. I think we should push @danieldeer to merge that PR.

mattsva: I did not read #26. That's nice, I've just saw it and it looks good

danieldeer: @crydafan @mattsva #26 will go in ASAP. It's waiting a rebase + removal of a committed `.sdb` file.

Gu11e: I could do it with `seek()` using a fixed amount of bytes while searching for the key, but I searched a bit and found a library called `ijson` which I could use instead.

Or maybe it's better to use the methods implemented in #26.

Honestly, I do not have a lot of experience working on open-source projects, so I'll stick to what you recommend!

main.py
Gu11e: The main reason I made it retrieve the content was for the header to include the right size of what would be the body, because FastAPI cuts the body and leaves the calculated size in the header, but I could instead read the size inside the funtion and return the response. 

That would still do the  
```python
val = db.get(key, None)
```
I don't know if there is any way to made the header include the size without loading `val`.

Please let me know which approach you would prefer.

mattsva:
Looks better

danieldeer:
@Gu11e @mattsva we can read from the in-memory cache now instead of opening the file. That would save even more lookup overhead. No need to open the file.

mattsva:
@Gu11e if you updated it, it would be good.

danieldeer:
Waiting for conflict resolution

main.py
```diff
@@ -76,15 +76,6 @@ async def get(key: str):
...
-        raise HTTPException(
-            status_code=503,
-            detail=f"Database file {db_file} could not be opened and loaded",
-        )
-    val = db.get(key, None)
-    if val is None:
-        raise HTTPException(status_code=404, detail=f"No value set for key {key}")
-    return val
+    val = select(key, cache)
+    size = len(val)
```
danieldeer: size is unused. Why?

Gu11e:
Sorry for the wait! It should work now. @danieldeer @mattsva

Gu11e:
@danieldeer I updated to the latest changes, you should be able to merge now.

crydafan (approved):
looks good to me!

AnnoDomine (approved):

PR #13, opened 2026-09-12
Refactor SDB file handling and ignore database file

TynK-M:
## Changes

- Removed redundant `f.close() calls`, as the `with ... as` context manager handles closing files automatically.
- Abstracted SDB read and write operations into `read_sdb()` and `dump_sdb()` helper functions.
- Added `.sdb` to `.gitignore` so the local database file is not committed to the repository.

## Notes

This keeps the database file handling cleaner and prevents the local .sdb database from being tracked by Git.

PR #38, opened 2026-09-13
Added .db and .sdb ignorance

mattsva:
Just a quick fix, that no one ever again will commit there .sdb files. (I added .db too just in case)


TynK-M:
It's already part of https://github.com/danieldeer/seriousdb/pull/13

mattsva:
> It's already part of #13


Yeah, but that would be two separate changes in one PR. I think that's better kept separate, especially since this needs to get into the main branch as quickly as possible so nobody accidentally commits those files. (I know "as quickly as possible" sounds a bit dramatic)

Don't get me wrong, your PR looks good, but I still think these should be two separate PRs. This one has a single purpose: adding the `.gitignore`. Keeping it separate makes the change easier to review and merge independently.

crydafan:
Hey, @mattsva while I agree that we need to make PRs scoped to a targeted issue, this adds too much bureaucracy to the project. Sometimes multiple changes in a single PR (like ignore tracking `.db` and `.sdb` files) are fine.

mattsva:
Yeah, I just want that to be merged, if yours get merge too, if it's good, mine isn't needed anymore and I close it without hesitation.


So if yours merge before.

mattsva:
Yeah, I maybe I close this. @danieldeer will for sure merge something in with the gitignore.

PR #42, opened 2026-09-13
Removing non-needed calls to close

Airbag65:
Since the file is opened within the `with` statements, the files are automatically closed when the execution leaves the scope of the `with` statement. The extra calls to f.close() don't do any harm at the moment but they are not needed and can thus cause unnecessary overhead


danieldeer:
Merged in a separate PR #43 

Next time I need to properly process the PRs, I messed up by essentially copying the provided code and merging it in my own PR. Sorry. The commit is still authored by you!

PR #43, opened 2026-09-13
Removing non-needed calls to close

danieldeer:
Since the file is opened within the `with` statements, the files are automatically closed when the execution leaves the scope of the `with` statement. The extra calls to f.close() don't do any harm at the moment but they are not needed and can thus cause unnecessary overhead

Contributed by https://github.com/Airbag65/, rebased by me.

Somehow github doesn't recognize that I modified the pull request in place


danieldeer:
@Airbag65 this is your PR, I have rebased it

PR #44, opened 2026-09-13
fix: Use 503 HTTP error code when database is unavailable

NotNekodev:
Previously 404 was returned. The Mozilla HTTP response code documentation reserves 4xx errors for client errors. Since it is not the clients fault that the database file is unavailable, we should use 5xx error codes (Specifically 503, which means "Service Unavailable")

This PR returns 503 instead of 404 if the database file cannot be opened

Source: https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/503


danieldeer:
Easy one liner & agree. Thanks!

PR #27, opened 2026-09-12
Add API test suite and concurrency regression coverage

FriedAxons:
## Summary

This PR adds an automated test suite for the API and regression coverage for concurrent database access.

### Testing

* Added `pytest` as a development dependency.
* Added FastAPI `TestClient` and pytest configuration.
* Added a fixture that gives each test its own temporary JSON `.sdb` database so tests do not modify the real database.
* Added tests covering:

  * `PUT /db`
  * `GET /db`
  * `GET /db/all`
  * Missing required API parameters
  * Missing-key behavior
  * Empty-key validation
  * Updating existing keys
* Added concurrent `PUT` requests to verify that multiple writers can access the database without losing data or corrupting the database state.
* Added concurrent `PUT` and `DELETE` requests to verify that different write operations remain consistent when performed simultaneously.

### Documentation

* Added `docs/testing.md` with instructions for running the test suite locally.
* Added the testing documentation to the README.

### Project Tooling

* Added Ruff as a development dependency to align with the project's current tooling.
* Configured pytest to discover the package from the `src` directory.

The test suite contains 19 tests and has been updated to work with the project's current JSON/cache and modular structure.

PR #46, opened 2026-09-13
feat: add key expiration

MarkVDD:
## Why

  Temporary data such as sessions, cached values, and idempotency keys should not need manual cleanup. TTL support lets these keys expire automatically without changing how permanent keys work.

  ## Implementation

  `PUT /db` now accepts an optional positive `ttl_seconds` value. Expired keys return `404` and are removed when read.

  Expiration times are stored in a versioned JSON format, so they survive restarts. Existing database files remain compatible and migrate on their next write.

  ## Validation

  Verified legacy reads, TTL writes, expiration, invalid TTL values, cleanup, and the existing `503` behavior using a temporary database.

  Tests are not included because #27 is currently establishing the project’s test setup.


danieldeer:
@mattsva any opinion on this? I mean the idea is not bad, I don't know if it will complicate our lives at such an early stage in the project.

MarkVDD:
I tried to keep the first version small: TTL is optional, expired keys are removed only when read, and there are no background tasks or new dependencies.

The main tradeoff is the versioned storage format needed to persist expiration times. Existing files remain readable and migrate on their next write. 

If that feels too early while the storage work is still evolving, I’m also happy to wait or adjust the implementation.

NotNekodev:
LGTM, but as said in [Daniel's Comment](https://github.com/danieldeer/seriousdb/pull/46#issuecomment-5652702835) I personally think it may be to early in the project to implement a feature like TTL on entries

main.py
```diff
@@ -1,44 +1,83 @@
...
 import os
-from fastapi import FastAPI
-from fastapi import HTTPException
+import time
+from typing import Annotated, TypedDict, cast
+
+from fastapi import FastAPI, HTTPException, Query
 
 db_file = ".sdb"
+db_format_version = 1
```
mattsva, line 9: DB format version is needed, but another PR probably. You may can leave it here if this gets merged, but thats needed at all.

docs/persistence.md
```diff
@@ -2,17 +2,33 @@
...
 
-```python
-{"default": "default"}
+```json
+{
+  "version": 1,
+  "entries": {
+    "default": {
+      "value": "default",
+      "expires_at": null
```
mattsva, line 14: The problem here is that it makes more data, a lot more.

main.py
```diff
@@ -1,44 +1,83 @@
...
+class Entry(TypedDict):
+    value: str
+    expires_at: float | None
+
+
+Database = dict[str, Entry]
+
+
+def default_db() -> Database:
+    return {"default": {"value": "default", "expires_at": None}}
```
mattsva, line 21: I think making it optional is definitely a good idea, no objections there. But wouldn't it be possible to make the overhead of storing the expiration information effectively zero when `"expires_at": None`? This might require some changes to the database structure itself, though.

mattsva (commented):
Doesn't this interfere with https://github.com/danieldeer/seriousdb/pull/26?

It is a big overhead, I am not sure if its needed right now. It looks about right, but I did not tested it on my machine in that current state.

mattsva:
> @mattsva any opinion on this? I mean the idea is not bad, I don't know if it will complicate our lives at such an early stage in the project.

It might be to early. I would go back to this at a stage where we have more of a real code structure, multiple files, and with that more of an option to implement such features, maybe we do another branch, and split it at some point in stable and experimental or something, that would make sense, wouldn't it? (not yet, but at all) @danieldeer

danieldeer:
Yes too early @mattsva. Let the project evolve a bit, and later we can add such features. But let's get something solid up first

MarkVDD:
That makes sense. I’ll close this for now so the storage design can settle first. The implementation can be revisited later if TTL fits the project’s direction. Thanks for the feedback!

PR #45, opened 2026-09-13
Update documentation for ruff usage

shoshta73:
closes #39

PR #47, opened 2026-09-13
chore: configure black target version

Max00009:
## Why
- Running `black --check .` on newer Python environments (such as Python 3.14+) raises AST safety check warnings when no explicit target version is defined in the configuration.
- Since the project specifies `requires-python = ">=3.11"`, setting Black's target version to `py311` resolves these warnings and ensures consistent formatting across different contributor setups.

## Implementation
- Added `[tool.black]` section with `target-version=['py311']` to `pyproject.toml`.

## Verification
- Ran `uv run black --check .` and verified that the check passes cleanly without warnings.

Before:
image

After:
image


mattsva:
I guess well dont use black anymore, so this isnt needed.

Look here: https://github.com/danieldeer/seriousdb/pull/45

AnnoDomine:
Closed (obsolete; project migrated from Black to Ruff).

PR #49, opened 2026-09-13
Feat: add docker

notcr12dev:
### Description

Add Docker support to make seriousdb easier to run and deploy in a consistent environment.

### Proposed changes

* Add a Dockerfile to build the seriousdb image.
* Add a `.dockerignore` file.
* Document Docker build and run instructions in the development.md

### Motivation

Docker support would simplify local setup, improve environment reproducibility, and make it easier to deploy seriousdb on servers.


danieldeer:
Agreed, good idea. I assume pip is natural to use inside the Dockerfile (as opposed to uv) because the base image is python-based. Adding uv would be unnecessary overhead (?)

Anyways, I am merging

mattsva:
> Agreed, good idea. I assume pip is natural to use inside the Dockerfile (as opposed to uv) because the base image is python-based. Adding uv would be unnecessary overhead (?)
> 
> Anyways, I am merging

Yeah, for `python:3.11-slim` `pip` is the least overhead.

PR #53, opened 2026-09-13
added a Delete endpoint that works with json

xqill275:
This is a new version of my delete endpoint, which now works with JSON.

Like last time, it takes a key as input and removes the key and its value from the database. It then returns:
`{
  "key": "test",
  "deleted_value": "dummy data"
}`


danieldeer:
@xqill275 can you adapt the delete endpoint to use the cache rather than updating the file directly? see the recent change contributed through #26

xqill275:
> @xqill275 can you adapt the delete endpoint to use the cache rather than updating the file directly? see the recent change contributed through #26

Done! let me know what you think

danieldeer:
Almost good. Maybe can you move the cache specific methods into the cache class, such that the actual delete_endpoint can be renamed "delete" without conflict?

After that I'd merge.

xqill275:
Your wish is my command @danieldeer

danieldeer:
Thr project restructured in the meantime. Could you also make your changes compatible with the latest updates? The cache class lives in a dedicated file now

xqill275:
should be good to go now

guy-gross:
Shouldn't a DELETE return status code 204, without any body? 
[Mozilla Dev - HTTP 204 No Content](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/204)

AnnoDomine:
Already implemented. Closed.

Issue #50, opened 2026-09-13
Project structure


danieldeer:
Almost all files are in root.

Is there some nice, standardized structure for such Python projects that we can use? Anyone have suggestions? I actually thought uv would create a src/ dir. It didn't.

NotNekodev:
After reading up a bit I think a project structure similar to this is good:
```
seriousdb/
├── docs/
│   ├── ...
├── seriousdb/
│   ├── main.py
├── tests/
│   ├── ...
├── LICENSE
├── pyproject.toml
└── README.md
``` 

While we obviously don't have tests at the moment, this would make it easy to organize and manage them

ShohinEshonov:
There are examples of project structure in the official FastAPI [documentation](https://fastapi.tiangolo.com/tutorial/bigger-applications/#an-example-file-structure)

However, I think that structure is a bit too complicated for the current stage of the project. I think we could start with something simpler, like:
```
seriousdb/
├── src/
│  └── seriousdb/
│       ├── __init__.py
│       ├── main.py
│       └── ...
├── tests/
├── docs/
├── pyproject.toml
├── README.md
└── ...
```
Then we could introduce additional modules or subpackages as the project grows.

danieldeer:
> ```
> seriousdb/
> ├── src/
> │  └── seriousdb/
> │       ├── __init__.py
> │       ├── main.py
> │       └── ...
> ├── tests/
> ├── docs/
> ├── pyproject.toml
> ├── README.md
> └── ...
> ```
> 
> Then we could introduce additional modules or subpackages as the project grows.

This looks good. Does anyone want to implement this?

ShohinEshonov:
I can work on this

danieldeer:
Okay, go ahead @ShohinEshonov. Thanks!

SrFluff:
I think we could see improvements in maintainability if we adopted a more portable structure, rather than having one `main.py` file. It would certainly reduce merge conflicts.

danieldeer:
Refactored

PR #56, opened 2026-09-13
Project structure improvment

NasriAnis:
## Summary
Refactors the monolithic `main.py` into a proper `app/` package for better separation of concerns, and fixes the `Dockerfile` to match the new structure.

## Changes
- **`app/cache.py`** `Cache` class plus `load()`/`flush()` for reading and persisting the JSON-backed store
- **`app/db.py`** `insert()`/`select()` key-value operations, decoupled from the FastAPI app setup
- **`app/main.py`** FastAPI app instance and route definitions only uses `Depends()` for cache injection instead of a bare module-level global
- **`app/config.py`** centralizes the `.sdb` file path
- **`run.py`** convenience entrypoint at project root (`python run.py`)
- **`Dockerfile`** updated `COPY` paths and `CMD` to target `app.main:app` instead of the old flat `main.py`

## Why
The original single-file `main.py` mixed storage, business logic, and route handling together, making it harder to test or extend. Splitting it up:
- Isolates the cache/persistence layer from HTTP concerns
- Makes `db.py` logic testable without needing FastAPI in the loop
- Replaces the module-level `cache` global with a dependency-injected one, so tests can swap in a fake cache later

## Testing
Now we can test by: `uv run run.py`

## Notes
- No behavior change to the public API routes, request/response shapes and status codes are identical to before.


danieldeer:
I had assigned the project structure improvement here https://github.com/danieldeer/seriousdb/issues/50, so likely this won't make it in

danieldeer:
But I VERY MUCH like the splitting of the code logic into db, cache, etc. 
We definitely need that. Just in a different folder structure. Let me await the project structure change first

NasriAnis:
Updated the structure to use `src/seriousdb/` as discussed this should align with what you had planned in #50. Let me know if this works or if you want it organized differently.

pyproject.toml
```diff
@@ -8,3 +12,6 @@ dependencies = [
     "black>=26.5.1",
     "fastapi[standard]>=0.141.1",
 ]
+
+[tool.hatch.build.targets.wheel]
```
danieldeer: What's this? I don't think this is needed, is it?

danieldeer:
I don't get these "hatchling" changes

NasriAnis:
Do you suspect me of using AI? These are just some files movement and `pyproject.toml` with dockerfile setup for new project structure.

Let me know if anything's unclear and I can walk through it in more detail.

danieldeer:
Yes, please explain the purpose of this "hatchlings" addition to the pyproject.toml file.
I don't suspect anything and I don't care where the code comes from, it should just be explainable

Dockerfile
```diff
@@ -3,10 +3,10 @@ FROM python:3.11-slim
 WORKDIR /app
 
 COPY pyproject.toml uv.lock* ./
-RUN pip install --no-cache-dir .
+COPY src/ ./src/
```
mattsva: This is not part of this PR. This is a different change ig.

run.py
```diff
@@ -0,0 +1,4 @@
+import uvicorn
```
mattsva, line 1: this is cleaner. Yes.

pyproject.toml
```diff
@@ -8,3 +12,6 @@ dependencies = [
...
+packages = ["src/seriousdb"]
```
mattsva: is it necessary to say `src/seriousdb` instead of just `src`?

mattsva (changes requested):
The structure isnt that good, but it is necessary to change it.

PS: I dont go in the debate with "hatchling" thats soemthing between @danieldeer and @NasriAnis.

pyproject.toml
NasriAnis: Yes with hatchling's src-layout convention packages = ["src/seriousdb"] tells it to take that directory and expose it as a top-level package named seriousdb (it strips the src/ prefix for the import name).

NasriAnis:
> Yes, please explain the purpose of this "hatchlings" addition to the pyproject.toml file. I don't suspect anything and I don't care where the code comes from, it should just be explainable

"hatchling" here refers to the build backend specified in pyproject.toml, declared under [build-system]:
```
toml
[build-system]
requires = ["hatchling"]
build-backend = "hatchling.build"
```

This tells Python's packaging tools (pip, build, uv, etc...) how to actually build the project into an installable package (wheel/sdist). Every pyproject.toml-based project needs one defined and without it tools don't know how to build from source.

If you find this useless it doesn't matter i can remove it.

danieldeer:
Now that hatchlings are removed, are we lacking any functionality that we would need?

danieldeer:
Again I need to emphasize here that I am not critical of you or the change. It's just my personal lack of knowledge in this area of Python. So I ask many questions @NasriAnis

NasriAnis:
> Again I need to emphasize here that I am not critical of you or the change. It's just my personal lack of knowledge in this area of Python. So I ask many questions @NasriAnis

There is absolutely no problem about that bro @danieldeer , its an opportunity for all of us to lean and i find that for now simplicity is better since we dont need any build system.

NasriAnis:
> Now that hatchlings are removed, are we lacking any functionality that we would need?

i think there is a lot of functionality are lacking for this project but for this pr only the structure modification are in scope.

NasriAnis:
> > Now that hatchlings are removed, are we lacking any functionality that we would need?
> 
> i think there is a lot of functionality are lacking for this project but for this pr only the structure modification are in scope.

i wanna ask only one thing what should i place there inside `pyproject.toml` (idk if you have write privileges or not for this pr):
```
description = "Add your description here"
```

danieldeer:
Okay, I'll merge if you tell me it's good @NasriAnis @mattsva

NasriAnis:
> Okay, I'll merge if you tell me it's good @NasriAnis @mattsva

yes its good to go.

danieldeer:
> > > Now that hatchlings are removed, are we lacking any functionality that we would need?
> > 
> > 
> > i think there is a lot of functionality are lacking for this project but for this pr only the structure modification are in scope.
> 
> i wanna ask only one thing what should i place there inside `pyproject.toml` (idk if you have write privileges or not for this pr):
> 
> ```
> description = "Add your description here"
> ```

Maybe "A key-value store over HTTP"? @NasriAnis

danieldeer:
Don't worry about the description for now. PR looked good overall so it goes in @NasriAnis . Thank you very much for this contribution

NasriAnis:
You are welcome @danieldeer, iam happy to be part of this project.

also note that in the README.md change `uv run fastapi dev main.py` to `uv run run.py`

pyproject.toml
mattsva: Sounds good

mattsva:
Greate outcome here!

PR #57, opened 2026-09-13
Make it better

HackingRepo:


danieldeer:
I will not accept contributions without description. "Make it better" is not sufficient

HackingRepo:
ok i will say, the pull request is to make the seriousdb to be synchrounous because async can make some bugs and also making error messages clear

danieldeer:
Closed. Very little reasoning. "can make some bugs" is not justification. Please properly describe PRs. If possible, open an issue first and discuss it

PR #48, opened 2026-09-13
Implement atomic writing to the database file

NotNekodev:
This pull request implements atomic writing to the .sdb file.<br>
- It first creates a temporary file and writes the database into  #that
- Flushes and synchronizes the temporary file to disk
- Replaces the temporary file with the actual one (atomically)
- Flushes the parent directory so processes can see the changes
 <br>
Implements #20

Issue #20, opened 2026-09-12
Make database writes safer


nikhil-thorat:
Currently, the database file is overwritten directly when a value is updated. If the process is interrupted during a write, the `.sdb` file could potentially be left in a corrupted or partially-written state.

It would be useful to make database writes atomic so that the existing database remains intact until the new data has been successfully written.

Would this be a change you'd be open to? If so, I'd be happy to work on a PR for it.

AnnoDomine:
As the way how the db file is managed changed with the project shift, this issue is legacy.
Closed.

PR #58, opened 2026-09-13
Implement atomic writing to the database file

NotNekodev:
Copy of #48 because GitHub was being weird with the merge conflicts.

Original description:
This pull request implements atomic writing to the .sdb file.

    It first creates a temporary file and writes the database into #that
    Flushes and synchronizes the temporary file to disk
    Replaces the temporary file with the actual one (atomically)
    Flushes the parent directory so processes can see the changes

Implements #20


danieldeer:
What's the benefit of the temp file, if we already lock on the writes to the cache?

main.py
```diff
@@ -12,25 +14,31 @@ def __init__(self):
 
 
 def insert(key: str, value: str, cache: Cache):
-    with cache.lock:  
+    with cache.lock:
         if cache.db is None:
-            raise HTTPException(status_code=404, detail=f"Database file {cache.filename} could not be opened and loaded")
+            raise HTTPException(
+                status_code=404,
```
Andre0n: I believe that using a 500 status code here is the best approach, since it's an internal error

LisZLisowni2:
Atomic writing sounds good on paper, but for now is probably unnecessary. Current storage layer handles operations synchronously, so there will not be any benefits from it now. Even slightly slowing I/O.

mattsva:
> What's the benefit of the temp file, if we already lock on the writes to the cache?

It is good for redundancy. It's just another layer of safety, in my opinion. Maybe we could make it optional, but I think the caching approach is a bit more advanced and probably the best fit here.

@LisZLisowni2 exactly.

NotNekodev:
> What's the benefit of the temp file, if we already lock on the writes to the cache?

The cache lock only protects against multi-threading inside the application. For example if an external program on the system modifies `.sdb` while writing it can cause pretty major corruption.

This also protects against program crashes which may happen during a write because the original `.sdb` is available until the last possible moment

TL;DR: The current cache lock is good for application-internal locking while the atomic writes help to secure it from factors outside of seriousdb

CC: @danieldeer @LisZLisowni2

LisZLisowni2:
I understand the concern around crash consistency and protecting against partial writes during system crashes.

However, forcing this logic introduces a heavy I/O overhead on every single request. External modification risks should be solved by file locking rather than atomic replaces.

In my opinion, the best compromise here is to make it optional via configuration, as @mattsva suggested.

mattsva:
I would suggest making for now a config variable in the code, that we can later change to a config file, if we have a plan there.
@NotNekodev so please do not implement a configfile here.

AnnoDomine:
Closing your PR does not mean it is rejected. We currently cleaning up to gain a better overview. Duplicate of #97 
Additional information: We are first need to decide in which way the API contract take place.

PR #60, opened 2026-09-13
Accept PUT values in request body

MarkVDD:
## Why

`PUT /db` currently sends both the key and value in the query string. The key identifies the entry, while the value is the data being stored, so accepting the value as JSON makes the request clearer and avoids encoding it into the URL.

## What changed

- Keep `key` as the query parameter used to identify the entry
- Require the value in a JSON body: `{"value": "Alice"}`
- Update the API and architecture documentation

Existing clients that send `value` as a query parameter will need to switch to the JSON body. The stored key-value data and persistence format are unchanged.

## Validation

- Verified JSON `PUT` returns `200` and the value can be read back with `GET`
- Verified the previous query-only form and a missing value return `422`
- Verified OpenAPI marks the JSON request body as required


docs/api.md
```diff
@@ -8,16 +8,21 @@ Interactive OpenAPI documentation is available at `http://127.0.0.1:8000/docs` w
...
 For example:
 
-```text
-key: name
-value: Alice
+```http
+PUT /db?key=name
+Content-Type: application/json
+
+{"value": "Alice"}
```
mattsva, line 25: No real overhead, makes the system better, but maybe harder to debug that easily the the Swagger UI, does it work there as convenient as before?

(Swagger UI is findable at `http://127.0.0.1:8000/docs` if the server is running)

I havent tested it yet. @MarkVDD could you?

docs/architecture.md
```diff
@@ -11,6 +11,7 @@ The service starts with a default entry when `.sdb` does not exist. There is no
 ## Request flow
 
 1. FastAPI receives a request.
-2. The route loads the dictionary from `.sdb`.
-3. A `PUT` updates and rewrites the file; a `GET` reads the requested value.
-4. The route returns the value or a `404` error.
+2. A `PUT` reads the value from its JSON request body.
```
mattsva, line 14: The loading of the DB is missing here.

main.py
```diff
@@ -12,25 +16,31 @@ def __init__(self):
...
         cache.db[key] = value
     return value
 
 
 def select(key: str, cache: Cache):
-    with cache.lock: 
+    with cache.lock:
         if cache.db is None:
-            raise HTTPException(status_code=404, detail=f"Database file {cache.filename} could not be opened and loaded")
+            raise HTTPException(
```
mattsva: This is formatting, not in this PR please.

Its OK for now tho ig.

mattsva (changes requested):
This is a good idea, needs to be updated tho (newest version of the repos main branch)

main.py
MarkVDD: Yeah, I thought of separating it, but figured it is small so not too bad to do here. If you'd rather have it in a separate PR, I don't mind doing that.

mattsva: I dont think its so necessary, but it helps to know what you actually changes.
Next time, only the changes said, such things as wrong formatting shouldnt be in the repo at all too.

docs/api.md
MarkVDD: image

Confirmed. In Swagger UI, key remains a query parameter and value is a required JSON request-body field. Sending {"value": "Alice"} returns 200 and the value can be retrieved with GET.

docs/architecture.md
MarkVDD: Added 👍🏼

docs/api.md
mattsva: Looks great, I see no problem.

mattsva (approved):
lgtm

AnnoDomine:
Thank you for your pull request!

As the project has matured and our vision for its direction has solidified, this proposed change is unfortunately no longer aligned with our current roadmap and has become legacy. Therefore, we are closing this PR.

We truly appreciate your effort and would love to invite you to stay involved with the project! You can find more details regarding our future direction here:
https://github.com/danieldeer/seriousdb/discussions/214

Thank you again, and we look forward to your future contributions!

PR #66, opened 2026-09-13
Update Dockerfile for new structure

notcr12dev:
# Description

- Updated the Dockerfile to a more scalable version to accommodate the new structure. 
- Additionally, `pip `has been removed and replaced with `uv`.


mattsva:
Since we're already using uv throughout the project, I'm fine with using it for dependency installation here.

One thing I'm wondering about though: do we need uv at runtime as well? Could we install the dependencies during the build and run uvicorn directly in the container instead of using uv run?

PR #69, opened 2026-09-13
feat(auth-template): A very simple template of how auth could be implemented

Airbag65:
### Foreword
This is a copy of #15 modified to fit the new and improved structure of the codebase

# Authorization
It might be a good idea to not let just anyone hit the endpoints as they please. This adds a very simple working template for how a future, more robust authentication system might look.

### Usage
- If you don't want you use any authentication at all, just toggle the `require_auth` flag in `auth.py` to false. If you do want authentication, toggle it to true.
- You'll need to set an `auth_token` in a `.env` (this PR comes with an example of that)
- In the request, use the `Autorization` header with the typical `Bearer <token>` pattern

If the token is not correct or the usage is not correct, you'll receive a `401 Unauthorized` response. Otherwise the endpoints will respond with what they usually do.

### NOTE
This is just a working POC and does not cover all the use cases that might be needed. I would be glad to continue working on a more finished and functional version of authentication, but I figured it might be best to create a POC before venturing too deep


.env.example
```diff
@@ -0,0 +1 @@
+AUTH_TOKEN=
```
mattsva: I would do it something like that:

```.env
AUTH_TOKEN=PLEASE_CHANGE_ME_FOR_PRODUCTION
```

Airbag65: Very logical. On it!

mattsva:
Isnt this "just a working POC". So this shouldnt be a Open PR in my opinion. It does have conflicts and is just a POC so its not open for a merge at this point.

PR #70, opened 2026-09-13
docs: correct Ruff contribution commands

MarkVDD:
## Why

The contribution documentation still tells contributors to run Black, even though the project uses Ruff as its development dependency now. The development guide also uses `ruff .`, which is not a valid Ruff formatting command AFAIK.

## What changed

- Replace the stale Black instruction with Ruff formatting and linting commands
- Document syncing the development dependency group before running those checks
- Correct the development guide to use `ruff format .`
- Add the non-mutating `ruff format --check .` command

No application behavior or dependencies are changed.

## Validation

- Ran `git diff --check`
- Verified Black is not declared in `pyproject.toml` or `flake.nix`
- Verified the updated branch contains no remaining Black references


mattsva (approved):
lgtm

fe-neu:
lgtm

PR #72, opened 2026-09-13
Recover from corrupt database file on startup

fedora-png:
'load()' crashes with 'JSONDecodeError' if 'sdb' exists but contains
invalid JSON. Since ' load() ' runs in the fastAPI startup handler, the
server never binds to a port. a single bad file takes the whole service
down with no recovery path.


crydafan (approved):
Looks good to me!

danieldeer:
look OK to me, merging

Issue #52, opened 2026-09-13
Move the database engine to a compiled language?


CosmicalBit:
Rn both the HTTP layer and the database logic are written in Python.

Would it make sense long-term to keep FastAPI/Python for the HTTP API, but implement the actual storage engine in a compiled language like Rust, C, C++ or zig?

That could make the database significantly faster and allow for lower-level optimizations as the project grows.

Obviously this would be a big architectural change. I’m mainly asking whether this is something you plan to do in the future.

mattsva:
Personally, I think that's a good approach. Something like C would probably be better in terms of performance, but I guess the use of Python was a decision made by @danieldeer from the beginning, so that's probably not something we should change now.

I would consider using custom Python packages with parts implemented in C or CPython when the time is right. For now, though, I think it's simply too early for that.

Good approach tho.

UniquePython:
Imo, this project is way too much in its early stages to consider that switch. I think once the project reaches a stage where all its major functionality and behaviour is nailed down, and, after repeated use and profiling there are blatant performance issues, then we can consider switching to a compiled language and thinking about optimizations

LisZLisowni2:
I think this is a good idea, but not for now. This is too early to write C/C++ or Rust modules of database management. But generally a good approach!

CosmicalBit:
> Imo, this project is way too much in its early stages to consider that switch. I think once the project reaches a stage where all its major functionality and behaviour is nailed down, and, after repeated use and profiling there are blatant performance issues, then we can consider switching to a compiled language and thinking about optimizations

Im suggesting it now bcs since it's early its easy todo. And most databases like  PostgreSQL, MySQL, SQLite RockDB etc are in compiled languages bcs performance always becomes critical

LisZLisowni2:
> > Imo, this project is way too much in its early stages to consider that switch. I think once the project reaches a stage where all its major functionality and behaviour is nailed down, and, after repeated use and profiling there are blatant performance issues, then we can consider switching to a compiled language and thinking about optimizations
> 
> Im suggesting it now bcs since it's early its easy todo. And most databases like PostgreSQL, MySQL, SQLite RockDB etc are in compiled languages bcs performance always becomes critical

This project is currently on prototype version, basic functions and architecture is still crystallising. If we do it now, it will be over-engineering, when we don't have clear architecture. Later when project will be functional we could catch performance bottlenecks and then rewrite database engine in compiled language.

mattsva:
[...]

PR #77, opened 2026-09-13
Basic implementation of a Rust db engine

CosmicalBit:
Added a basic implementation for a rust db engine. i enabled some settings witch i like and are my recommendation but feel free to change it. its the suggested engine in   #52


CosmicalBit:
Maybe it could still be worth keeping the compiled-engine experiment on a separate branch/fork for now. That way it doesn’t affect the current direction, but once the database is more developed we could compare the two approaches with actual workloads and see whether it’s worth revisiting

VihaanReddyM:
I agree with @CosmicalBit, I dont think we are going to towards rust rn

AnnoDomine:
@CosmicalBit 
I will close this PR as the implementation is at the moment not relevant.
But i like including a dedicated db-engine.
I would prefer you should open a discussion or an issue to discuss this on a dedicated board :)
You can link this PR there too.

CosmicalBit:
> @CosmicalBit 
> 
> I will close this PR as the implementation is at the moment not relevant.
> 
> But i like including a dedicated db-engine.
> 
> I would prefer you should open a discussion or an issue to discuss this on a dedicated board :)
> 
> You can link this PR there too.

@AnnoDomine  I did that before the Pr. He mentioned my Pr in the most recent vídeo. He said he didn't want a dedicated db in other language, it says it's not worth it

PR #80, opened 2026-09-13
Auto-fixes using `uv tool run ruff check . --fix`

danieldeer:


danieldeer:
Should be obvious. I simply ran the ruff linter for the first time using `uv tool run ruff check .` as per documentation, and auto-applied fixes as per `--fix` flag.

PR #81, opened 2026-09-13
feat(static-config): Adding static config object

### LINKED ISSUE #134: Static app configuration
I guess this is an issue to make the discussion in the [dicsussion thread](https://github.com/danieldeer/seriousdb/discussions/41) offifcal. 

TLDR;
To begin with, we'll have a JSON based configuration that is serialized into a dataclass object. 

See discussion tread linked above to see more

Airbag65:
Closes #134 

Adding a static configuration system. Config is primarily defined in a JSON file that is then being serialized into a `dataclass`. If the file is completely missing, a fallback with sane defaults will be used. Also, if one or more fields is/are missing from the config, default values will be used for those field.

If a field that is not recognized by the serialization is encountered, the program will still run and use the rest of the fields that were valid are used, but the program warns about it and prints which fields are allowed. 

I also added a bunch of unit-tests for the configuration system using the de facto test framework for python: `pytest`. Also I added `.github/workflows/test.yml` to automatically run `pytest` on push.

I welcome all feedback! :smile:  If you find anything that would obviously benefit from being made in a different way, I'll gladly make some modifications.


AnnoDomine:
Hey @Airbag65,

Thanks for putting work into the configuration setup!

I went through the implementation in configuration.py. While the fallback handling to dataclass defaults is nice, the whole setup feels unnecessarily complex:
- Managing config.json in the root and test.config.json forces file-based configuration.
- We end up with custom reflection, type checking, and warning classes just to parse a JSON file.
- In Docker, serverless, or CI environments, configuring via environment variables is much more natural than mounting a JSON file.

Wouldn't it be much simpler to use plain environment variables with in-code fallbacks? For anyone who still wants a local file, PR #151 already brings in dotenv support.

What do you think about dropping the JSON file logic and keeping it simple with environment variables?

PR #82, opened 2026-09-13
Sync uv lock file and use dependency injection

crydafan:
Just a simple PR. The use of `typing.Annotated` is because Ruff was complaining about singleton default arguments. Following FastAPI dependency injection convention, I wrapped the cache dependency with `typing.Annotated`.


crydafan:
@danieldeer would you mind taking a look at this? Thank you!

Issue #64, opened 2026-09-13
Rewrite of entire database file for every PUT request


Max00009:
Current code rewrites the database file each time a new PUT request arrives.

What if we do a cheap append-only writes on the hot path or WAL(Write Ahead Log) instead of full-file rewrites every time.And then after a fixed interval we do compaction with atomic write safety.
This way the database file is rewritten only during compaction.
If the program crashes suddenly we can still recover uncompacted changes from WAL file and save it to the database file in next boot.

can raise a PR if needed

mattsva:
Could you make a implementation/POC of this, with all new features included. If its actually better, it gets merged.

Max00009:
Okay

mattsva:
> Okay

I may were unclear, I would do that in your position. Im srry for being unclear.

crydafan:
Redis has a similar mechanism called AOF (Append-Only File) that logs every write operation received by the server as it happens.

What you propose is actually a good and cheap solution to our flush bottleneck problem https://github.com/danieldeer/seriousdb/pull/26#discussion_r3997306869 as it eliminates the complexity of an asynchronous solution like I mentioned.

Feel free to open a PR with these changes.

Max00009:
Okay I am working on it

Max00009:
https://github.com/danieldeer/seriousdb/pull/97

AnnoDomine:
As the db file handling is changin with the project shift, this issue is legacy.

PR #84, opened 2026-09-13
Make flush a background task

crydafan:
Before this patch the request would wait for a disk flush before sending back the response. Now, by using [background tasks](https://fastapi.tiangolo.com/tutorial/background-tasks/#using-backgroundtasks) the flush operation is offloaded.

Per my comment in a previous PR https://github.com/danieldeer/seriousdb/pull/26#discussion_r3997306869 a good solution would be to offload the flush task to a worker thread, of course, this also had its issues, namely concurrent writes and possible file corruption because the database file is rewritten on every PUT request. Now that we have mitigated the concurrent writes (file corruption was a possibility from the get-go) it is safe to offload the flush.

There are multiple ways to mitigate database corruption. See https://github.com/danieldeer/seriousdb/issues/64 for one solution.


crydafan:
@danieldeer

src/seriousdb/main.py
acai10, line 1: Could we make this endpoint synchronous since it does not await anything and both insert and flush are synchronous? Also, since flush runs as a background task, the response may be returned before the data is persisted, and any flush failure cannot be reported to the client. Is that the intended consistency guarantee for this endpoint?

crydafan: Leaving `async` in the endpoints guarantees those handlers are run in the event loop, thus removing the overhead of spinning up a new thread every time a client makes a request to the server.

Thank you for spotting the issue with background tasks, I forgot to mention it in the description of the PR. I think it boils down to consistency vs availability. Flush failures won't be reported to the client with this approach. Flush failure is something that must be hardened in the database engine with other approaches. See https://github.com/danieldeer/seriousdb/pull/97.

acai10: Keeping the endpoints asynchronous makes FastAPI execute them directly on the event loop and avoids dispatching them to a worker thread. But FastAPI uses a thread pool and does not create a new thread for every request. Since these handlers call synchronous code and use a synchronous lock, defining them with def may be safer to avoid blocking the event loop. Or am I wrong?

crydafan: I don't think you're wrong. Regarding the `async def` and `def` discussion there was a PR that replaced the former with the latter.

danieldeer:
Awesome contribution. Thanks a lot. Approved and merged.

PR #86, opened 2026-09-13
chore: use multi-stage build

argoarsiks:
The image was shipping `uv` and the whole build toolchain, none of
which is needed once dependencies are installed. They now live in a
builder stage, and only `/app/.venv` is copied into a plain
`python:3.11-slim-bookworm`.

| | size | compressed |
|---|---:|---:|
| **before** | 450 MB | 108 MB |
| **after** | **296 MB** | **69.4 MB** |


AnnoDomine:
@argoarsiks 
This only affects docker user? Or i am wrong?

argoarsiks:
> @argoarsiks This only affects docker user? Or i am wrong?

Yeah, just for Docker

PR #91, opened 2026-09-13
feat(nix): migrate to flake-parts, add docker derivation, and add e2e VM tests

snqre:
1. update flake to use flake parts framework which is scalable and readable

2. update dev environment - nix/dev/default.nix (new update adds git pre commit hooks for nix users checking for conventional commits compliance, size (no accidental large files), accidental secrets exposure, toml, json, and configuration checks)

3. move `run.py` inside `seriousdb` package (tooling expects this to be within the package) (tooling can point to the run.py at the root, but this is typically an anti pattern. as the project grows, there may be several packages in src, `run.py` becomes ambiguous as to what it is actually running)

5. add nix derivation for `seriousdb` - nix/pkg/default.nix

7. add nix docker image derivation for `seriousdb` - nix/pkg/default.nix (NOTE: nix docker images are much leaner and efficient as they know exactly all the dependencies that need to be built into the final image, making them much smaller and faster to build)

9. add nix e2e sanity check - nix/check/sanity.nix (run with nix flake check or example nix build .#checks.x86_64-linux.sanity) (you can run e2e tests and simulate complex networking topologies (for example several machines using the same database at the same time), to stress test the databse)

10. fix warnings and type errors - src/seriousdb/cache.py

new commands available
- `nix run` - default command to run `seriousdb`
- `nix run .#seriousdb`
- `nix run .#seriousdb_docker_image`
- `nix build` - default command to build `seriousdb`
- `nix build .#seriousdb`
- `nix build .#seriousdb_docker_image`
- `nix flake check .#checks.x86_64-linux.sanity` - .#<checks.<host>.<check_name>>

you can load the docker image from `result` which is the `nix` target

step(0): `nix build .#seriousdb_docker_image`
step(1): `./result | docker load`
step(2): `docker run -p 8000:8000 seriousdb:latest`


GabrielTessier:
Your modifications for nix look good. But you made changes that have nothing to do with nix, like moving a python file or adding types. I think you should split your changes into different issues/pr

snqre:
thank you for the feedback @GabrielTessier, i realize there are a few moving parts here, but these changes directly support one another (you can view it as an infrastructure upgrade)

1. standard python packaging tools and nix derivations rely on package entry points living inside the module namespace (`src/seriousdb`). leaving `run.py` at the root creates module resolution ambiguities depending on execution path (`sys.path`) and forces custom workarounds in `nix/pkg/default.nix`. moving it inside ensures our `nix` app and `docker` image builds are strictly reproducible. leaving the entry point at root, is unnecessary technical debt and works against idiomatic `uv` and tooling usage

2. root-level entry points don't scale when a repository expands to house multiple packages or binaries. keeping entry points scoped to their respective package directory prevents ambiguity

3. the typing updates were required to pass the `nix` flake checks (`nix flake check`) introduced in this pr so that ci pipelines build without warnings. as they now check and capture these warnings and errors. (*these can be stripped and removed from the pr is required*) (i haven't seen a serious project come to completion without proper typing, especially when collaboration is amongst people is required, these also act as documentation)

these changes directly enable the new `nix build` targets and `flake check` pipelines, keeping them in this pr ensures the new nix setup works out of the box

removing any of these would deliver this pr in a partially broken state, wether it be containing warnings or errors. with exception to the movement of `run.py` into its respective package (which can be moved back if required, but has no technical reason for living there)

.pre-commit-config.yaml
```diff
@@ -0,0 +1 @@
+/nix/store/x70hnq5cf6wmh1612c9jf914j6gpngx5-pre-commit-config.json
```
MarkVDD, line 1: This is specific to your machine AFAIK, please remove it :)

GabrielTessier: No it's not specific to his machine. It's always the same path because the version is set in flake.lock
If you do `nix develop` you will see that you have the same path and the symlink is ok.

I just want to know why it's a `.yaml` because the original file is a json file not a yaml file ?

snqre: @GabrielTessier is correct, in `nix` store paths are generated using a cryptographic hash of the input source and build inputs

on evaluation, the same inputs and outputs return the same hash string (in this case `x70hnq5cf6wmh1612c9jf914j6gpngx5`) on every machine

regarding `.yaml`, it just happens to be the selected format for `pre-commit`, which is the tool that is hooked into `nix` git hooks. i can only speculate as to the reason they have chosen `.yaml` is because git hooks are already largely in `.yaml`, and they want to keep the convention

by all relevant metrics, both `.json` and `.yaml` represent the same information exactly the same

MarkVDD: @snqre @GabrielTessier Ah, you're right, thanks for the explanation. Looks good to me then :)

GabrielTessier:
For e2e tests, i think it's better to use the python script inside the tests folder

snqre:
`nix` e2e tests are much more thorough and powerful than the samples seen in `tests`, you can simulate large complex topologies via `nix`, these operate in isolation, which make them more reliable than `docker`

these containers simulate entire machine networks down to their individual system configuration

the `tests` directory may contain focused local checks, but when you need production grade assurance that the system works under real load, those will show up in `nix/check`

i think it's best to leave both on the table, this will open up the option for more powerful orchestration in the future

if they decide to use it or not is up to them, deleting the `check` folder is as simple as submitting another PR, i just want them to know it can be done and it is a powerful tool in the arsenal

AnnoDomine:
This PR includes a lot of side parts which are not relevant to nix.
Please reopen a new one with only the changes which are relevant for nix and not affect users which are not using nix.

PR #95, opened 2026-09-13
Add function return type annotations

OskarEek:
Add function return type annotations for the following functions to improve readability:

In db.py:
- insert
- select

In cache.py:
- _write_default


AnnoDomine:
Already implemented.

Issue #90, opened 2026-09-13
Data Loss / File Corruption on Write Interruption (CWE-372)


AnnoDomine:
**Issue:**
Inside `src/seriousdb/cache.py` (lines 45–50 and 34–41), inside the `flush` function, the database file is written directly using `open(cache.filename, "wb+")`. With the `"wb+"` argument, the file is immediately truncated before the JSON bytes are written.

If the process dies (for example due to a system shutdown, SIGKILL, or Out-Of-Memory error) during the write process, the file will be left with partial/corrupted JSON.

When the app restarts, `load()` catches `json.JSONDecodeError` / `UnicodeDecodeError`, moves the `.sdb` file to `.sdb.corrupt-<timestamp>`, and resets the database back to the default state (`{"default": "default"}`), resulting in total loss of stored database data.

**Suggested fix:**
Write to a temporary file first and atomically replace the `.sdb` file using `os.replace`:

```python
import tempfile
import os

def flush(cache: Cache):
    with cache.lock:
        if cache.db is None or cache.filename is None:
            return
        dir_name = os.path.dirname(cache.filename) or "."
        with tempfile.NamedTemporaryFile("wb", dir=dir_name, delete=False) as tf:
            tf.write(json.dumps(cache.db).encode())
            temp_path = tf.name
        os.replace(temp_path, cache.filename)
```

**Proof of Concept (PoC):**
1. Populate the database with data (e.g., ~5,000 key-value pairs).
2. Send a request to the `PUT /db` endpoint with a new key and a large value.
3. While the write is in progress, forcibly kill the application process (e.g., `CTRL+C` or `kill -9 <pid>`).
4. Restart the application.
5. **Result:** The application reports a corrupt database file, moves `.sdb` to `.sdb.corrupt-*`, and resets all stored data back to `{"default": "default"}`.

**References:**
[MITRE CWE-372](https://cwe.mitre.org/data/definitions/372.html)

AnnoDomine:
I saw another PR opend #125

danieldeer:
I am working on integrating this code change. It is a major improvement, thanks @AnnoDomine.
I will add also a `fsync` call to ensure the data is physically stored. As I understand, regular flush and close only guarantee the data is moved from application level to OS-level. But there's no guarantee about physical storage yet. `fsync` promises that

danieldeer:
Implemented

PR #97, opened 2026-09-13
feat: add WAL(Write-Ahead Logging)-based durability, atomic compaction, and thread-safety to Db_Engine

### LINKED ISSUE #21: Load the DB into memory at startup instead of on every request
Every GET and PUT opens `.sdb` and unpickles the whole thing, so reading one key deserializes the entire dataset. The data already ends up in RAM on every request, it just gets thrown away right after.

Would you take a change that loads the dict once at startup (in a lifespan handler) and keeps it there? Reads would hit memory directly, writes would still go through to disk, so nothing changes on durability.

Once the dict is the source of truth, a `SERIOUSDB_PERSIST=0` for a disk-less mode is easy to add on top, which would also help with tests.

Can send a PR if you're interested.

Max00009:
## Why
- Right now the code rewrites the whole database for each PUT request which is redundant.Instead we can load the changes directly on memory and after a fixed interval we compact current memory into database file.
- What if the program crashes before compact is triggered?->We will maintain a temporary,cheap WAL(Write Ahead Logging) file.We open that file in append mode and everytime a PUT is called we push that pair in that file.And when next compact is triggered we just flush that file empty and start writting again.This way if our program crashes before compacting ,We can just look if there is any WAL file available with contents during next bootup and load that into memory.And next compact will write that into the database file.
- this way we don't loose any data and also we don't have to rewrite the sdb file for each PUT request.

## Changes
- db_engine.py replaces cache.py and db.py. db_engine.py will handle boot,put,get and compact functionality with multithreading safety.
- main.py is cleaner only handle PUT and GET requests .also async is removed to have concurrency with thread pool.
- config.py contains more config values.
- .gitignore contains runtime generated files

## Tests
<b> Test 1: Server Running </b>
- First  I tested if the server is running successfully by running `uv run run.py`
image

---

<b> Test 2: .sdb file generated </b>
- It also generated the .sdb file immediately with default values:
image

---

<b> Test 3: GET and PUT working </b>
- We can successfully send and receive data:
image
image

---

<b> Test 4:WAL file (changes.log) file generated </b>
- Upon sending the first PUT request we can see our data is stored into the changes.log file:
image

- We can also see that .sdb file still has only the default values cause DATABASE FILE WILL BE REWRITTEN ONLY AFTER COMPACTION which is after fixed interval(50 put request in my environmant)
image

---

<b> Test 5: Database file rewrite after compaction </b>
- To simulate compaction I sent 49 more put requests cause that will add up to 50 which is COMPACT_THRESHOLD value in my environment.So it should trigger compact and write all 50 entries to .sdb file (total 51 including the default value).And it should leave changes.log file empty.

- sending 49 put requests:
image

- changes.log is empty and .sdb file contains 51 entries after compaction
image

---

<b> Test 6 : Persistent changes.log after program crashes </b>
- First I put some data in change.log to show that it will persist after crash:
image

image

- Next,to simulate a program crash I first kill the server process and then the reloader process:
image

image

- After crashing the program we can still see the entry is available in changes.log(WAL file):
image

---

<b> Test 7: Recovering the uncompacted data from WAL file </b>
- Now when I run the program again it should load the WAL file into memory and rewrite that into disk at next compaction.
- Restarting the server:
image

- Sending 49 more put requests to trigger next compaction:
image

- After compaction, the data before crash is written into .sdb file:
image

- No data loss.

---

<b> Test 8: Manual compaction during normal server shut down </b>
- When we manually shut down the server it triggeres the compaction irrespective of how much data is inside WAL file.It will rewrite the memory into .sdb file before shutting down.
- shutting down the server manually:
image

- All contents of WAL file is saved into .sdb file:
image

- WAL file is empty now:
image

Fixes #21


src/seriousdb/db_engine.py
```diff
@@ -0,0 +1,125 @@
...
+            # step1: Update memory with new key
+            self.state[key] = value
+
+            # setp2: Append to WAL file
+            with open(WAL_FILE, "a") as f:
+                f.write(
+                    json.dumps({key: value}) + "\n"
+                )  # using dumps instead of dump cause we are working with string.not file object
+                """below two lines are to make sure that the OS actually write to the WAL file before returning success to the client.
+                Without os.fsync(), calling f.write() only puts data into operating system memory (page cache).
```
lexxmach: this explanation is partially wrong.

f.write() MIGHT put your data into kernel-space memory
but It'll try to fill it's internal buffer before invoking real syscall

that's why flush is needed beforehand, and also AFAIK fsync also waits for the actual disk to apply it's internal caches onto the real storage.

src/seriousdb/db_engine.py
```diff
@@ -0,0 +1,125 @@
...
+                with open(WAL_FILE, "r") as f:
+                    for line in f:
+                        if line.strip():
+                            entry = json.loads(
+                                line
+                            )  # loads is being used instead of load cause we are loading string(each line in WAL file) not a file object
+                            self.state.update(entry)
+
+                # call compact to merge with DB_FILE
+                self._compact_locked()
```
lexxmach: This needs better exception handling.

If we write onto the WAL file, and then (even with WAL_SYNC_ON_WRITE) linux goes off/db crashes, we might see only partial line here.

This line was actually never seen in the real DB state, nor does user got 200 from put request. So it's safe to assume, that this line is garbage, and we can just skip it, rather than failing to boot

src/seriousdb/db_engine.py
Max00009: Yeah, you are right.I oversimplified the process and skipped one step.I had f.write() reaching the OS page cache directly, but that's not quite accurate.It first writes to python's own internal buffer ,then flush()  pushes to  OS page cache, via the write() syscall.
and yeah fsync() asks the disk's onboard cache to write into physical memory and waits for it to complete.

I have fixed the comment.

Thanks for the correction.

MarkVDD (commented):
I personally think the implementation is over-commented. Many comments narrate individual lines or repeat what the code already says, which makes the core logic harder to scan and maintain, again, my opinion, unsure how others feel about this.

orstro (commented):
Nice direction — the WAL + atomic-compaction approach is exactly the right fix for #90/#92, and the fsync-on-write handling is thoughtful. A few things I'd flag before this is mergeable:

**1. Existing test suite currently fails against this engine (verified).**
The e2e tests in `tests/test_e2e.py` isolate the DB by mutating `main.DB_FILE` to a temp path, but `db_engine.py` binds `DB_FILE` at import time (`from .config import DB_FILE`), so the engine ignores the patch and writes a real `.sdb` into the CWD. `test_put_persists_to_db_file_on_disk` reads the temp file that never gets written and `FileNotFoundError`s. I ran the suite against this branch to confirm.
- Suggest injecting the paths into `Db_Engine` (e.g. `DbEngine(db_file=..., wal_file=...)`) instead of reading module constants, and/or having tests pass a temp dir in.
- Nobody else has touched them, but this PR adds no tests for the new WAL/compaction logic — that's the highest-risk code in the repo now. At minimum: (a) boot replays a WAL with real entries, (b) compaction truncates and persists, (c) a torn/crash-interrupted WAL file.

**2. A corrupt/torn WAL line hard-crashes startup.**
`boot()` replays the WAL line-by-line with a bare `json.loads(line)`. A crash mid-append before fsync can leave a partial final line (exactly what fsync-mode is meant to survive), and `json.loads` then raises `JSONDecodeError` *outside* the recovery block, so the server fails to start. The old code had backup-and-fresh recovery for DB_FILE but this path has none. Suggest catching parse errors and either backing up the WAL or trimming the trailing partial line (defaulting to the 'fresh' branch).

**3. Rebasing needed — currently dirty.**
The PR is based on 0b00be5, but main has since merged the background-flush change (#84) and the e2e test suite + CI (#105), so the change overlaps and `mergeable_state` is dirty. Once rebased, point #1 becomes CI-breaking.

**Minor:**
- For full crash-safety the `os.replace` should also fsync the parent directory, otherwise the rename itself can be lost after a sudden power loss.
- `Db_Engine` -> `DbEngine` (PEP 8). And on the over-commenting note from MarkVDD's review, agreed — the WAL section in `put()` is a good docstring candidate rather than a long inline block.

Max00009:
Thank you for your feedback.
I have made changes:
 1. I have enabled injecting paths into DbEngine with validation check. Reading module constants is fallback now.
 2. I have added separate test file 'test_db_engine.py' to test for DbEngine's WAL, compaction, crash-recovery, and validation logic etc.Also changed the `test_put_persists_to_db_file_on_disk` test according to new architecture.All tests pass now.
 3. Added logic to safely handle JSONDecodeError in case of loading corrupt/torn WAL line.Now it stops when a torn line is encountered and skips everything after it.
 4. Added _fsync_dir() function call inside _compact_locked() and boot() for full crash safety.It will fsync() the parent directory after os.replace()
 5. Rebased.
 6. Chnaged class name Db_Engine->DbEngine.
 7. Moved the inline block WAL section into docstring.

AnnoDomine:
@Max00009 
Could you please merge the current main to your branch?

Max00009:
@AnnoDomine  Yeah working on it

amin-xiv:
@orstro did you really have to use AI to write this?

AnnoDomine:
> @orstro did you really have to use AI to write this?

Hey.
Thanks for your critic. But a PR commentary history is not realy the place for such comments.

AnnoDomine:
@Max00009 
There are a lot of merge conflicts which are highly complex.
Would it be possible to rebase your main and cherry pick the commits step by step and create multiple PRs for a better overview?
From my perspectiv every change is highly valid and improves the project without a lot of complexity.
Optional you can reach out to me on discord, so we can pair resolve this together.

For now to have a better overview of the PRs i marked it as draft.

Max00009:
@AnnoDomine I am trying to fix all conflicts before merging the main branch with my feat branch.But it's taking some time and sometimes I am just confused.

And thank you.It would be really helpful if I can discuss with you on discord.

AnnoDomine:
> @AnnoDomine I am trying to fix all conflicts before merging the main branch with my feat branch.But it's taking some time and sometimes I am just confused.
> 
> And thank you.It would be really helpful if I can discuss with you on discord.

Believe me. It is way more easy to cherry pick the single commits than to try to resolve everything :)
Otherwise, When do you have time tomorrow?
The Discord link you can find in the Discussion: https://github.com/danieldeer/seriousdb/discussions/108

AnnoDomine:
@Max00009 As we communicated, Splitting the implementations into multiple atomic PRs would be better to not lost the overview.
And as a red teamer we already communicate, these are really good improvements which we would like to discuss in a seperated Issue.
I will now close the PR:

Thank you for your pull request!

As the project has matured and our vision for its direction has solidified, this proposed change is unfortunately no longer aligned with our current roadmap and has become legacy. Therefore, we are closing this PR.

We truly appreciate your effort and would love to invite you to stay involved with the project! You can find more details regarding our future direction here:
https://github.com/danieldeer/seriousdb/discussions/214

Thank you again, and we look forward to your future contributions!

Max00009:
@AnnoDomine Thanks for guiding me.😄
and yeah I would love to contribute to the project's success.

PR #99, opened 2026-09-13
feat: add TTL-based key expiration

### LINKED ISSUE #79: TTL (Time-To-Live) / Key expiration and Strict types
Right now database only stores data permanently, but what if we add special parameter called TTL (Time-To-Live), which tells us how long certain key have to live? I propose checking that every get operation (Lazy Expiration) for start and passing optional parameter in put method.

What do you think? 
I could open a PR pretty soon.

orstro:
Resolves #79

Adds an optional `ttl` query parameter to `PUT /db`. When provided, the key
expires after that many seconds.

**Implementation**
- `insert` stores entries with a TTL as `{"value": ..., "expires_at": <epoch>}`;
  plain entries stay plain strings, so existing `.sdb` files remain valid.
- `select` lazily checks expiration on every GET and removes expired keys,
  returning `404`.
- `ttl` is validated by FastAPI (`gt=0`, `422` otherwise).
- Docs updated (`docs/api.md`, `docs/persistence.md`).

**Verified**
- `ruff check` / `ruff format --check` clean.
- Live server test: PUT with ttl=1 returns value immediately, 404 after expiry;
  non-TTL key unaffected; ttl=0 returns 422.


mattsva:
This is really similar to https://github.com/danieldeer/seriousdb/pull/46. As we said there, this is a overhead and on this stage of the project not needed. It maybe something for later version(s), maybe making this optional.

NullSec8:
Thanks — I hadn't seen #46. I agree, let's let the storage design settle first. The pytest/TestClient setup from this branch is reusable, so I'll close the PR and can revisit TTL once the codebase is more solid. Thanks for the pointer!

PR #105, opened 2026-09-14
Add E2E sanity tests + CI for documented API

### LINKED ISSUE #103: Sanity / E2E testing on CI
just an idea, what if we add a CI to test the documented features before it gets merged to main?

danielshamsudin:
## Summary
- Adds `tests/test_e2e.py`: sanity tests against the documented HTTP API (`docs/api.md`, `docs/persistence.md`) — default-seed key, PUT/GET round-trip, overwrite, 404 on missing key, and on-disk persistence.
- Adds `.github/workflows/e2e.yml`: runs these tests on every PR and push to `main`, using `uv sync` + `uv run python -m unittest discover -s tests`.
- No new dependencies: `TestClient` only needs `httpx`, which is already pulled in transitively by the existing `fastapi[standard]` dependency.

Closes #103.

## Test plan
- [x] `uv sync` from a clean checkout
- [x] `uv run python -m unittest discover -s tests -v` — 5/5 passing locally
- [x] Verified no stray `.sdb` file is created in the repo (tests use an isolated temp dir)


danieldeer:
Thanks a lot. Approved!

PR #107, opened 2026-09-14
feat: add health readiness endpoint

### LINKED ISSUE #106: Add health/readiness endpoint
The application currently has no dedicated health or readiness endpoint. Deployment systems, container orchestrators, and CI checks cannot reliably determine whether the FastAPI process has completed its startup lifecycle and is ready to serve database requests.

The application loads the database synchronously in FastAPI's lifespan handler before normal requests are accepted. A readiness endpoint provides a stable operational signal after that startup work completes, rather than requiring infrastructure to infer readiness from a normal database request or process status.

## Implemented solution

- Added `GET /health` to the FastAPI application.
- Readiness is based on whether the shared in-memory database cache completed startup loading.
- A ready service returns `{"status": "ok"}` with HTTP `200`.
- An unavailable cache returns HTTP `503 Service Unavailable` with a stable, non-sensitive response. This is defensive behavior for later cache-state failures or lifecycle changes; if startup fails before requests are accepted, the endpoint cannot respond.
- Added focused tests for the real startup lifecycle and the defensive unavailable-cache branch without relying on the repository's real `.sdb` file.
- Documented the endpoint in the API reference and development documentation.
- Addressed review feedback by adding cleanup-safe shared-state isolation and platform-safe `pathlib.Path` handling. Pytest migration remains a separate repository-wide tooling decision.

## Example

```python
@app.get("/health")
def health(cache: Annotated[Cache, Depends(get_cache)]):
    if cache.db is None:
        raise HTTPException(status_code=503, detail="Service unavailable")
    return {"status": "ok"}
```

## Benefits

- Gives Docker, CI, and deployment systems a stable readiness check.
- Signals that the FastAPI startup lifecycle has completed and the database cache is available.
- Avoids exposing filesystem paths or internal database errors to monitoring clients.
- Establishes a foundation for future liveness, metrics, and observability improvements.

## Acceptance criteria

- [x] `GET /health` is added to the FastAPI application.
- [x] A real startup lifecycle with an isolated database returns HTTP `200` and `{"status": "ok"}`.
- [x] An unavailable cache returns HTTP `503` without exposing internal details.
- [x] Tests cover both the startup-ready path and the defensive unavailable-cache path.
- [x] API and development documentation describe the endpoint.
- [x] Existing database endpoints continue to behave as before, verified by the full 7-test suite.

tonytrieu-dev:
## Summary

Closes #106

- add `GET /health` readiness endpoint
- return `200 {"status":"ok"}` when the database cache is loaded
- return `503` with a stable, non-sensitive response when the cache is unavailable
- add focused tests for the real startup lifecycle and unavailable-cache state
- document the endpoint in the API and development guides

## Test quality

The two health tests cover the meaningful behavioral boundaries without redundant route permutations:

- the lifecycle test uses an isolated temporary database and `TestClient` as a context manager, proving that the real FastAPI startup handler loads the cache and makes `/health` return 200
- the unavailable-cache regression test uses dependency injection to exercise the defensive 503 branch and verify that internal details are not exposed

The lifecycle setup is a reusable cleanup-safe context manager that restores the shared cache state even when the test fails. It uses `pathlib.Path` for platform-safe temporary paths.

The 503 state is defensive coverage. Normal FastAPI startup completes before requests are accepted, so the endpoint cannot report a startup failure that prevents the server from serving requests. The branch remains important if cache state becomes unavailable later or the application lifecycle changes.

## Review follow-up

- Addressed the shared-state isolation and temporary-path feedback from review.
- Replied to all reviewers directly; the inline threads are resolved.
- Pytest migration is intentionally left for a repository-wide tooling decision rather than mixing a test-framework migration into this focused endpoint contribution.

## Validation

- `uv run python -m unittest discover -s tests -v` — 7 tests passing, including the current E2E suite
- `uv run ruff check tests/test_health.py src/seriousdb/main.py`
- `uv run ruff format --check tests/test_health.py src/seriousdb/main.py`

The full repository Ruff check/format check still reports the existing `src/seriousdb/cache.py` formatting issue from the current base branch, covered by open PR #101.

The branch now contains four focused commits: the endpoint implementation, the startup lifecycle test, the test-isolation improvement, and a merge of current `main` to resolve the application refactor conflict while preserving this feature.


orstro (approved):
Verified locally against the PR branch:
- `python -m unittest discover -s tests -v` -> 2 passed
- `ruff check` and `ruff format --check` clean on both changed files

Nice and minimal endpoint - matches the project's style. A few non-blocking suggestions:

1. **Test framework alignment:** this PR uses `unittest`, while the project direction (discussion #12, PR #27) is leaning toward pytest. Not blocking - and easy to align later once the maintainer settles on one.

2. **Global state mutation in the first test:** swapping `main.cache.filename/db` and restoring them in `finally` works, but it has a soft spot - if `client.get()` raises, `response` stays unbound and you'd get a `NameError` instead of the real failure (I'd initialize `response = None` before the block). Also, mutating the shared `main.cache` global can leak into other tests as the suite grows; the `dependency_overrides` approach used in the second test is safer and could be reused here.

3. **Cosmetic:** `f"{directory}/test.sdb"` could be `pathlib.Path(directory) / "test.sdb"`.

None of these block merging. Nice work, and the nod to #101 was appreciated. :)

tests/test_health.py
```diff
@@ -0,0 +1,47 @@
+import unittest
```
MarkVDD, line 1: The coverage looks right. Could we make this test’s temporary cache/database setup a small reusable fixture that restores both `main.cache.filename` and `main.cache.db` after each test? This keeps the shared application cache isolated, so future endpoint tests cannot accidentally inherit state from this health test.

MarkVDD (approved):

tests/test_health.py
acai10, line 1: This looks fine for now, but once pytest is introduced, it would be worth migrating these tests. Pytest provides simpler assertions, less boilerplate, reusable fixtures, automatic setup and cleanup, and helpful built-in fixtures such as tmp_path and monkeypatch. It also generally makes tests easier to read, extend, and maintain.

tests/test_health.py
tonytrieu-dev: Good catch. I took this suggestion: the lifecycle test now uses a small reusable context manager that snapshots and restores both `main.cache.filename` and `main.cache.db` in `finally`, so cleanup still happens if the request or startup raises. I also switched the temporary path construction to `pathlib.Path`. Thanks for pointing out the shared-state risk.

tests/test_health.py
tonytrieu-dev: I agree pytest would be a better fit once the repository settles on it, especially for reusable fixtures and clearer setup/cleanup. I’m leaving that as a separate tooling change rather than mixing a framework migration into this focused endpoint PR; the current tests stay aligned with the repository’s existing unittest-based command. I documented that scope decision in the PR body.

tonytrieu-dev:
@orstro Thanks for taking the time to verify the branch locally and for the thoughtful review. I addressed the shared-state concern by moving the lifecycle setup into a reusable cleanup-safe context manager that restores both cache fields even if startup or the request raises, and I switched the temporary path to `pathlib.Path`.

I’m keeping the tests on `unittest` for this PR so the focused endpoint change stays aligned with the repository’s current test command. I agree pytest would be a good direction once the project settles that tooling decision, and I’ve called that out in the PR description rather than mixing a framework migration into this change.

Thanks again for the approval and for catching the test-isolation improvement.

AnnoDomine:
Please resolve merge conflicts.
Can be merged after.
**Approved**

tonytrieu-dev:
@AnnoDomine Thanks for the approval and for flagging this. I’ve merged the latest `origin/main` into the contribution branch and resolved the conflicts while preserving the `/health` endpoint and the upstream API/error-handling changes.

I also updated the health regression assertion for the standardized `503` response shape. The full validation is now green: 39 pytest tests, 9 unittest-discovery tests, full Ruff check, and full Ruff format check. GitHub now reports the PR as clean and mergeable.

Issue #93, opened 2026-09-13
Use JSON Request Body Instead of Query Parameters for PUT Endpoint


AnnoDomine:
Currently, the application mutates database data by receiving `key` and `value` via URI query parameters (`PUT /db?key=...&value=...`).

These parameters should typically be sent in the request body as JSON (`PUT /db` with a JSON payload).

Using a JSON request body serves multiple purposes:
1. **REST Conventions:** Adheres to standard REST API design practices for data modification (`PUT`/`POST`).
2. **Log Privacy:** Prevents sensitive key and value data from being exposed in plain text inside server access logs, HTTP proxies, or browser histories (where query strings are logged by default).

**Suggested Fix:**
Define a Pydantic model for the request payload in `src/seriousdb/main.py`:

```python
from pydantic import BaseModel, Field

class ItemPayload(BaseModel):
    key: str = Field(..., min_length=1, description="Key name")
    value: str = Field(..., description="Value to store")

@app.put("/db")
def put(payload: ItemPayload, cache: Annotated[Cache, Depends(get_cache)]):
    insert(payload.key, payload.value, cache)
    flush(cache)
    return {"key": payload.key, "value": payload.value}
```

aunrxg:
Yes i noticed this on day one itself, but now since daniel is moving from HTTP to full fledge database is this still relevant as it was earlier?

PR #111, opened 2026-09-14
Add Request Body to put method and updated api.md

pastaepiselli:
#93 Hi this is my first pull request to an open source project.

I have modified put method to accept a json and updated the api.md.


acai10:
lgtm

MarkVDD (commented):
This overlaps with #60, which already changes PUT to accept JSON input and documents that contract. However, #93 proposes moving both key and value into the request body, while #60 deliberately keeps key as a query parameter.

We should first decide the intended API contract for #93, then keep one implementation rather than merge two competing PUT designs.

AnnoDomine:
Closing your PR does not mean it is rejected. We currently cleaning up to gain a better overview. Duplicate of #60

PR #120, opened 2026-09-14
feat: allow configuring db file path

EnricoDev1:
Adds support for configurable database path with `SERIOUSDB_FILE` environment variable.
By default, the application still uses `.sdb` as database file.

I added a small config test which verifies that `DB_FILE` uses the correct value.


AnnoDomine:
Closing your PR does not mean it is rejected. We are cleaning up to gain a better overview. Duplicate of #81

PR #124, opened 2026-09-14
Added test for delete endpoint

### LINKED ISSUE #122: Add test for new DELETE endpoint
A new delete endpoint was added.

No test covers it so far. Can anyone take care?

OskarEek:
Closes #122 

Not sure if the 2 get requests are really needed since for example `test_put_stores_value_and_get_retrieves_it` already tests the same steps. Im open for feedback on if I should remove them or not.


FriedAxons:
I noticed `tests/test_e2e.py` now includes coverage for the `DELETE /db` endpoint as well.

Just for context, PR #27 was intended as the broader API test suite and already includes dedicated tests for successful deletion, missing-key deletion, validation, missing parameters, updating values, and concurrent `PUT` / `DELETE` operations, along with the pytest configuration and testing documentation.

So there is some overlap with the new end-to-end sanity test, but #27 is intended to provide the more comprehensive regression suite, while `test_e2e.py` appears focused on a smaller set of documented end-to-end sanity checks.

Also, I noticed the delete test in `test_e2e.py` appears to assert against `put_response` rather than `delete_response`, so that may be worth correcting.

OskarEek:
@FriedAxons 
Fixed the issue with asserting against `put_response` rather than `delete_reponse`
You might want to suggest that your PR closes #122 ? This issue was just created by @danieldeer asking for tests for the DELETE-endpoint

FriedAxons:
> @FriedAxons Fixed the issue with asserting against `put_response` rather than `delete_reponse` You might want to suggest that your PR closes #122 ? This issue was just created by @danieldeer asking for tests for the DELETE-endpoint

Good suggestion. PR #27 already includes dedicated coverage for the `DELETE /db` endpoint as part of the broader API test suite, so I’ll reference `#122` in the PR so it can be closed when the test suite is merged.

AnnoDomine:
Closing your PR does not mean it is rejected. We currently cleaning up to gain a better overview. Duplicate of #27

PR #125, opened 2026-09-14
Write database atomically via temp file and os.replace

fedora-png:
flush() opened .sdb with "wb+", which truncates the file to 0 bytes on open  so a crash mid-write left an empty file and lost allstored keys. The fix writes to a temp file first, then `os.replace()`s it over .sdb atomically, so the real file is never in a partial state.


AnnoDomine:
FYI: #90 and #97

AnnoDomine:
Duplication from #58

danieldeer:
Implemented

PR #127, opened 2026-09-14
feat: support for different types in DB

### LINKED ISSUE #110: allow all json data types
the project would be more useful if it allows more data types. this could be implemented as `json.load(value)` in `insert`

VihaanReddyM:
## Summary

This PR adds support for storing JSON-compatible value types instead of restricting database values to strings.

### Changes

* Add a `JsonValue` type alias covering JSON-compatible Python values:

  * strings
  * integers and floats
  * booleans
  * `None`
  * lists
  * dictionaries
* Add a parser for converting the `value` query parameter from a string into the appropriate JSON/Python type.
* Keep JSON-looking strings representable by using JSON string syntax (for example, `"42"` remains a string while `42` becomes an integer).
* Update the API documentation with examples of the supported value formats.
* Update the relevant tests.

### Examples

```text
value=42
```

is stored as an integer.

```text
value="42"
```

is stored as the string `"42"`.

Complex JSON values are also supported:

```text
value=[1, 2, 3]
```

```text
value={"name": "Alice"}
```

### Status

Ready!


VihaanReddyM:
Integration tests were also added but lemme know if they look too simple I can add more

VihaanReddyM:
this was from #110 btw

VihaanReddyM:
Updated the PR to reflect new changes

AnnoDomine:
I am not confirm with this.
The meaning of this project is KEY-VALUE
With integrating of multy types value at the moment, there should be a lot of checks included like value type check, sanitization, etc.
This would be way to complex for a simple key-value list.

VihaanReddyM:
Will fix the conflicts soon

AnnoDomine:
@VihaanReddyM 
Multy types fields are an improvement. Sure.
But as we additional serve an API layer within this project, we have to take care about it too.
For this i would point you out to: https://blog.postman.com/rest-api-best-practices/
Spezific the section "Version your API deliberately" and the sub-section "When to create a new version".

VihaanReddyM:
I agree that we should consider API compatibility since SeriousDB exposes an HTTP API. However, I don't think introducing a new API version is necessary for this change.
The project is still in its initial development phase, and we're currently establishing the API contract rather than maintaining a long-standing, stable API. This PR adds support for JSON-compatible value types, which is a core database feature, while existing string values remain supported.
I think introducing v2 at this stage would add unnecessary API complexity for a feature that belongs in the initial API design. I would expect versioning to become appropriate once the API has stabilized and we need to introduce changes that genuinely cannot remain backward compatible.
I'm happy to make sure the existing string behavior is covered by tests so we don't unintentionally break existing clients.

AnnoDomine:
> I agree that we should consider API compatibility since SeriousDB exposes an HTTP API. However, I don't think introducing a new API version is necessary for this change. The project is still in its initial development phase, and we're currently establishing the API contract rather than maintaining a long-standing, stable API. This PR adds support for JSON-compatible value types, which is a core database feature, while existing string values remain supported. I think introducing v2 at this stage would add unnecessary API complexity for a feature that belongs in the initial API design. I would expect versioning to become appropriate once the API has stabilized and we need to introduce changes that genuinely cannot remain backward compatible. I'm happy to make sure the existing string behavior is covered by tests so we don't unintentionally break existing clients.

I think you missunderstand it.
We are not related to spezific type of a field inside the JSON. This users have to take care by there own.
It is the part about, if a type is changed between the inital definition and the expected response.
As an example:
- Client creates a field which is a number.
- API stores the field as a number.
- Client request the field and expect a number.
- API sends the field as a number.
- Anything happens and the field change it type from a number to a list.
- Client request the field and expect a number.
- API sends the field as a list.
- Client crash as the expected type is different to the responsed one.

VihaanReddyM:
@Airbag65 I think I am going to close this pr. thoughts?

Airbag65:
> @Airbag65 I think I am going to close this pr. thoughts?

Go ahead!

PR #28, opened 2026-09-12
Update API reference documentation

FriedAxons:
## Summary

Update the API reference to document the current API, including the additions and validation changes introduced in this PR.

* Add documentation for `GET /db/all`.
* Add documentation for `DELETE /db`.
* Add documentation for `HEAD /db`.
* Document that `PUT /db` requires a key with at least one character.
* Document the `422` response for an empty key.
* Document the `404` response when a requested key does not exist.
* Document the possible `503` response when the database has not been loaded.
* Add examples for the API en*

PR #131, opened 2026-09-14
Add latest API endpoints to API docs

CodeShady:
- Adds documentation to the existing `DELETE /db` and `GET /db/all` endpoint to the written API reference
- ~~Fixes an issue where deleted keys were removed from the in-memory cache but not persisted to the JSON database.~~

~~Currently, deleting a key removes it from the in-memory cache, but the change is not persisted to the database. This means restarting the server reloads the delete key back into the cache, making it accessible again.~~

~~No major changes, the persistence fix simply calls `cache.flush`.~~


VihaanReddyM (approved):
Looks good!

VihaanReddyM:
@CodeShady the deletion persistence bug was fixed please revert changes related to that and resolve the conflicts, thanks!

CodeShady:
@VihaanReddyM I'm a bit confused why the database persistence problem was fixed on its own rather than just merging my PR, but all good, I reverted changes and updated the API documentation to reflect the latest changes. :+1:

VihaanReddyM (approved):
Docs changes, looks good

AnnoDomine:
Already documented in #28 and flush is already implemented on main.

PR #119, opened 2026-09-14
feat: centralize exception handling with custom error responses

acai10:
Closes #96.

## Summary

Replaces per-function `HTTPException` checks with centralized application error handling.

- Adds an application error hierarchy with `ResourceNotFoundError` (`404`) and `ServiceUnavailableError` (`503`).
- Adds centralized FastAPI exception handlers with consistent error responses.
- Prevents unexpected `500` responses from exposing internal details.
- Introduces `cache.require_db()` as the shared database availability check.
- Removes FastAPI dependencies from the cache layer.
- Updates `api.md` and `architecture.md`.

This branch also includes the latest `main` changes, including the `Cache` method refactor, removal of `db.py`, and the new `DELETE /db` endpoint.

Database loading failures now return `503` instead of `500`. All other documented behavior remains unchanged.

The existing 5 E2E tests pass, and Ruff reports no issues. Additional error-handling tests will follow in a separate PR.

PR #137, opened 2026-09-14
feat: add centralized logging

### LINKED ISSUE #102: Introduce logging
The application currently has no logging at all. Errors, warnings, and important events are not recorded anywhere, which makes debugging issues and understanding production behavior difficult.

For example, cache/database failures currently only surface as an HTTP error response to the client — there's no record of what happened, when, or why. The same lack of visibility likely applies to other parts of the application.

We should introduce basic, consistent logging across the application.

## Proposed changes

- Add a central logging configuration (format, level, handlers), e.g. in `logging_config.py`, initialized at app startup.
- Use `logging.getLogger(__name__)` in modules that need to log, instead of `print()` or nothing.
- Make log level configurable via environment variable (e.g. `LOG_LEVEL`).
- Log unexpected exceptions with stack traces (e.g. in a global exception handler), without exposing internal details to clients.
- Use the existing cache/database error handling as the first use case.

## Example

```python
# logging_config.py
import logging, sys

def configure_logging(level: str = "INFO") -> None:
    logging.basicConfig(
        level=level,
        format="%(asctime)s %(levelname)s %(name)s %(message)s",
        stream=sys.stdout,
    )
```

```python
logger = logging.getLogger(__name__)

def require_db(cache: Cache) -> dict[str, str]:
    if cache.db is None:
        logger.warning("Database unavailable: %s", cache.filename)
        raise ServiceUnavailableError(...)
    return cache.db
```

## Benefits

- Visibility into errors and application behavior
- Easier debugging and incident investigation
- Foundation for future monitoring/alerting

## Acceptance criteria

- [ ] Central logging configuration is added and initialized at app startup.
- [ ] Key modules (starting with cache/database) log relevant events and errors.
- [ ] Log level is configurable via environment variable.
- [ ] Unexpected exceptions are logged with stack traces, without exposing internals to clients.
- [ ] Tests cover logging configuration where reasonable.

aditya-sutar-45:
Closes #102 

- Add centralized logging configuration
- Support configurable log levels through `LOG_LEVEL` from env
- Add cache/database logging for important events and errors
- Log unexpected exceptions with stack traces
- Return a generic 500 response for unexpected exceptions

- Tests `(test_logging.py)`: 9 tests covering logging config (default/custom/invalid levels), cache log events (unavailable DB, missing key, new DB, corrupt DB recovery), and the global exception handler (stack trace logged, internals not leaked).

### run the tests with
```
uv run python -m unittest discover tests -v
```

### change log levels with
```
LOG_LEVEL=DEBUG uv run python run.py
```


src/seriousdb/logging_config.py
acai10, line 1: If you are using `os.getenv()` I would suggest to introduce a `.env.example` and also use `from dotenv import load_dotenv`.

Edit: Since many use `os.getenv()` without using a `.env.example` I think this might be an issue for later, I will open one for the future dev environment. So in my opinion you can leave it as it is regarding the `getenv()`.

acai10: Another matter, duplicate `logger = logging.getLogger(__name__)` ? There is also one in `main.py`, one of them should be redundant or are they different?

acai10: Could we simplify this by using levels.get(level_name, logging.INFO) instead of tracking invalid_level separately? We can then check if level_name not in levels before logging the warning. This would keep the same behavior while making the function a little shorter and easier to read.

VihaanReddyM:
@aditya-sutar-45 I would like to help you out in this PR is that ok with you or would you like to work on it alone?

src/seriousdb/logging_config.py
aditya-sutar-45: Hello, 

I have made the suggested changes to simplify the log level configuration. 

As for the duplicate loggers,
The two logger instances are separate because `__name__` is different in each module (`seriousdb.main` and `seriousdb.cache`), so I kept them separate to identify where the logs are coming from.

EXAMPLE:
```
2026-09-15 11:01:45,980 INFO seriousdb.main THIS LOG IS FOR THE MAIN MODULE
2026-09-15 11:01:45,980 INFO seriousdb.cache Loaded database from .sdb
```

aditya-sutar-45:
> @aditya-sutar-45 I would like to help you out in this PR is that ok with you or would you like to work on it alone?

Sure @VihaanReddyM , I'd appreciate the help! 

Maybe you could take a look through the implementation and the tests, see if there's anything that could be improved or if there are any other places where logging would make sense

VihaanReddyM:
@aditya-sutar-45 I looked into the implementation, and I was thinking: why not use `config.py` to store the configuration since I think that would be preferable for now?

Currently the logging level is being fetched from the environment inside `configure_logging()`. Instead we could load and validate it once in `config.py` and then use the value from there.

```py
LOG_LEVEL = os.getenv("LOG_LEVEL", "INFO").upper()

if LOG_LEVEL not in logging.getLevelNamesMapping():
    LOG_LEVEL = "INFO"
```

aditya-sutar-45:
@VihaanReddyM Yeah, that makes sense.

One thing though, what would happen to the warning for an invalid `LOG_LEVEL`? Would we still want to log that somewhere?

VihaanReddyM:
> @VihaanReddyM Yeah, that makes sense.
> 
> 
> 
> One thing though, what would happen to the warning for an invalid `LOG_LEVEL`? Would we still want to log that somewhere?

We can add typing to log_level so that if we have an invalid log_level we send a Warning and default to Info log level. Warning can printed or there should be a workout. I am in class rn so will look into it and get back to you in sometime

AnnoDomine:
Overlaps with #119

PR #141, opened 2026-09-15
feat: Validate /db query params with Pydantic schemas

DaviAlcanfor:
The PUT/GET/DELETE endpoints took key/value as bare str args, with no structured validation. They now use Pydantic schemas via Query(), keeping the same URL shape (?key=&value=) and response format.

schemas.py (new):
```python
KeyField = Annotated[str, Field(description="...", examples=["name"])]

class KeyRequest(BaseModel):
    key: KeyField
    value: ValueField

class KeyQuery(BaseModel):
    key: KeyField
```
main.py:
```python
@app.put("/db", response_model=str)
def put(params: Annotated[KeyRequest, Query()], ...):
    cache.insert(params.key, params.value)
    ...
```
Existing tests (tests/test_e2e.py) pass unchanged.


src/seriousdb/schemas.py
```diff
@@ -0,0 +1,15 @@
...
+ValueField = Annotated[str, Field(description="The value associated with the key.", examples=["Alice"])]
+
+
+class KeyRequest(BaseModel):
+    key: KeyField
+    value: ValueField
+
+
+class KeyQuery(BaseModel):
+    key: KeyField
```
AnnoDomine, line 15: Nice and clean use of Annotated! Two small suggestions for improvement:

1 Inheritance (DRY):
Since both models share the key field, we can leverage Pydantic's model inheritance:
```suggestion
from typing import Annotated

from pydantic import BaseModel, Field

KeyField = Annotated[str, Field(description="The key associated with the value.", examples=["name"])]
ValueField = Annotated[str, Field(description="The value associated with the key.", examples=["Alice"])]

class KeySchema(BaseModel):
    key: KeyField
    
class KeyValueSchema(KeyModel):
    value: ValueField

```

2 Naming clarity (optional suggestion):
Renaming KeyRequest to KeyValueModel (or KeyValueRequest) would be more descriptive, as it explicitly communicates that this payload expects both the key and the value (and avoids confusion with pure query parameters).

src/seriousdb/main.py
```diff
@@ -22,23 +23,22 @@ def get_cache() -> Cache:
...
     background_tasks.add_task(cache.flush)
-    return value
+    return params.value
 
 
-@app.get("/db")
-def get(key: str, cache: Annotated[Cache, Depends(get_cache)]):
-    return cache.select(key)
+@app.get("/db", response_model=str)
+def get(params: Annotated[KeyQuery, Query()], cache: Annotated[Cache, Depends(get_cache)]):
```
AnnoDomine, line 38: What are your thoughts on creating a type alias for the cache dependency?

Instead of repeating cache: Annotated[Cache, Depends(get_cache)] in every endpoint, we could define a dependency alias at the top (which is also the official FastAPI best practice):
```py
CacheDep = Annotated[Cache, Depends(get_cache)]
```

Then the endpoints become much cleaner and easier to read:
```py
@app.get("/")
def get(params: Annotated[KeyQuery, Query()], cache: CacheDep):

#...

@app.delete("/")
def delete(params: Annotated[KeyQuery, Query()], cache: CacheDep):
    ...
```

AnnoDomine (commented):
Added comments and suggestions.
In additional FYI: #93

AnnoDomine:
Closing your PR does not mean it is rejected. We currently cleaning up to gain a better overview. Duplicate of #60 
Additional information: We have first to decide which API contract will take place.

src/seriousdb/main.py
DaviAlcanfor: That's great actually! I didn't think of that previously. Also, why not create an alias for the params?

src/seriousdb/schemas.py
DaviAlcanfor: Alright, thanks for the suggestion! I'll consider it.
You have any thoughts on any Response model? Does it need to return Key and Value or just Key?

DaviAlcanfor:
So do I still have the oportunity to have this or not?

AnnoDomine:
> So do I still have the oportunity to have this or not?

Your work is not lost Davi.
As you recognise by yourself the current state of the PRs is chaos.
For this we need first make more decissions and gain a better overview.
The PR 60 have a link in the list which points to your.
This means when we go over these, we will take a look into yours too :)

PR #145, opened 2026-09-15
Added tests for delete

oliverstalsberg22-png:
Covers regular deletes, deleting already deleted keys, and checking if deletes persist on the disk. Failed last two tests. Both checked whether data was deleted from disk

closes issue #122


AnnoDomine:
There are already a couple of open PRs adding tests for the DELETE endpoint, but I really like how granular and focused your test cases are instead of having one huge test that tests everything at once.

One (or two) small things :)
Take a look at your last two tests: `test_put_persists_in_db_after_delete` and `test_put_deleted_from_db_after_delte`.
Typos and contradiction.

AnnoDomine:
Closing your PR does not mean it is rejected. We currently cleaning up to gain a better overview. Duplicate of #27

Issue #144, opened 2026-09-15
A standalone CLI for server configuration management


LisZLisowni2:
As the project grows, more and more environment variables will be added. Currently there isn't validation for envs which can cause errors on server starting or unexpected behaviour of it. My proposal is a CLI separate from the main program (separate directory in src, example src/seriousdb-cli) that can safely manage the configuration. 

Commands (for now): 

- `seriousdb-cli config init` — Generates a default configuration template.
- `seriousdb-cli config show` — Displays the active configuration (with sensitive data masked).
- `seriousdb-cli config set <KEY> <VALUE>` — Safely updates specific settings.
- `seriousdb-cli config check` — Validates settings (e.g., valid port ranges, directory write permissions) before starting the main server.

Probably it won't require any extra dependencies if we use built-in argparse, however we can include typer in future.
Also that architecture (subcommands) will allow expand CLI in future

If that meet your vision, I'd be happy to be assigned to this issue and open a PR soon.

AnnoDomine:
Good idea and thoughtful proposal!

However, I think we should keep it simple (KISS) and focus on the core database functionality first before introducing additional tooling.
The project is currently at too early of a stage for a dedicated CLI. Introducing a standalone configuration tool right now would add unnecessary maintenance overhead, as any future configuration changes in the main app would always need to be mirrored and maintained in the CLI as well.
If the main goal is validating environment variables (like valid port ranges or permissions), handling that on startup directly within the app (e.g. via `pydantic-settings` or simple startup checks) would solve the problem much more lightweight without needing a separate tool.

LisZLisowni2:
It makes sense with CLI, we can revisit it in future. For now I can agree with you, for enviroment variables validation function in main program are enough.

Would it be okay if I work on PR for that startup validation instead?

AnnoDomine:
I haven't seen any open issue or PR covering environment variable validation on startup yet, so go for it!
Looking forward to your PR!
But beware... I will inspect it in detail 🧐

LisZLisowni2:
Thanks. I'll make sure it's top quality. I will open PR soon.

LisZLisowni2:
I have created a PR: #146. You can check up.

LisZLisowni2:
Since the PR #81 exist and #146 was closed, should we keep this issue open for potential CLI/tooling extensions in the future, or close it for now?

AnnoDomine:
[...]

PR #146, opened 2026-09-15
feat(config): add startup environment variable validation and .env parser

LisZLisowni2:
Adds validation for environment variables and .env parser without adding extra dependencies. Also includes tests and documentation for configuration using .env file. Refers to #144 issue


.gitignore
AnnoDomine, line 1: Missing `.env` in .gitignore

src/seriousdb/cache.py
```diff
@@ -2,7 +2,9 @@
 import logging
 import os
 import time
+from pathlib import Path
```
AnnoDomine: Can't see if Path is used here.

src/seriousdb/config.py
```diff
@@ -1 +1,42 @@
-DB_FILE = ".sdb"
+import os
+from pathlib import Path
+
+BASE_DIR = Path(__file__).resolve().parent.parent.parent
```
AnnoDomine: Path traversing can lead into a problem when the file is changing where it placed.

src/seriousdb/config.py
```diff
@@ -1 +1,42 @@
...
+
+            key, value = line.split("=", 1)
+            os.environ.setdefault(key.strip(), value.strip().strip("'\""))
+
+
+def validate_config(env_path: Path = ENV_FILE):
+    load_env_file(env_path)
+
+    # DB_FILE
+    DB_FILE_RAW = os.getenv("DB_FILE")
```
AnnoDomine: Can be done more easy instead of an if statement use the build in system `os.getenv("DB_FILE", ".sdb")`.

AnnoDomine (changes requested):
Reviewed the PR.
You can highly improve everything.
Just pointed to a few improvments.

.gitignore
LisZLisowni2: I looks like .env in already covered in .gitignore. Look at line 151.

image

AnnoDomine: Ah got it. I forgot daniel used the python template.

LisZLisowni2:
I have recently updated my code. Please review again my code.

AnnoDomine:
> I have recently updated my code. Please review again my code.

Hey Lis!

I went through the code again. While the implementation is solid, I think it is a bit over-engineered for our current needs:

- What happens if someone adds 10+ new env vars in future PRs?
- Do we really need to validate defined variables when sensible fallback defaults exist?
- How easy is this to maintain long-term as the codebase grows?

Additionally, we are currently triaging open PRs to clean up the backlog. Since this overlaps and conflicts with earlier config PR #81, I am going to close this for now to keep things consolidated.

Please note that closing this does not mean your work is rejected! I am always happy to give feedback or bounce ideas around anytime. Thanks for contributing!

LisZLisowni2:
Thanks for the review and clear explanation, @AnnoDomine. 
I fully understand your decision, I didn't see #81 PR before.
Looking forward to picking up a new issue (or if I come up with an idea, I will create a issue)

PR #150, opened 2026-09-15
Made deletes persist

### LINKED ISSUE #149: DELETE does not persist deletion, causing deleted keys to reappear after restart
## Description

Deleting a key through `DELETE /db` removes the key from the running instance, but the deletion is not persisted to disk.

As a result, the key appears to be deleted until the database is restarted, after which the previously deleted key is loaded back into the database.

### Steps to reproduce

1. Start the database.
2. Insert a key:

```bash
curl -X PUT "http://localhost:8000/db?key=name&value=Alice"
```

Response:

```text
"Alice"
```

3. Delete the key:

```bash
curl -X DELETE "http://localhost:8000/db?key=name"
```

Response:

```text
"Alice"
```

4. Verify that the key is deleted:

```bash
curl "http://localhost:8000/db?key=name"
```

Response:

```json
{"detail":"No value set for key name"}
```

5. Restart the database.
6. Query the same key again:

```bash
curl "http://localhost:8000/db?key=name"
```

Response:

```text
"Alice"
```

### Expected behavior

Once `DELETE /db` successfully deletes a key, that deletion should be persisted. After restarting the database, the deleted key should remain absent and the request should return `404`.

### Actual behavior

The key is removed from the running instance, but the persisted database still contains the key. After restarting, the key is loaded back into the database.

VihaanReddyM:
- Made deletes persist to the disk
- Fixed the Formatting

Fixes #149


Airbag65 (commented):
Simple and straight forward. LGTM

danieldeer:
can be merged as soon as conflicts are resolved

VihaanReddyM:
Conflicts have been resolved!

Issue #147, opened 2026-09-15
Add GET /db/keys endpoint to list stored keys


mohammedFawzy0111:
There's currently no way to discover what keys exist in the store without already knowing them. I'd like to add a GET /db/keys endpoint that returns all stored keys, with tests.

xqill275:
Hey, I made a pr for this: #148

mohammedFawzy0111:
Hey @xqill275  I read your pr and the comments. And I would be happy to discuss with you

mohammedFawzy0111:
We can implement the db method itself in the cache  class something like that 
```python
class Cache:
    ...

    def keys(
        self,
        limit: int,
        cursor: str | None = None,
    ) -> tuple[list[str], str | None]:
        ...
```
```python
@app.get("/keys")
def list_keys(
    limit: int = 100,
    cursor: str | None = None,
    cache: Cache = Depends(get_cache),
):
    keys, next_cursor = cache.keys(
        limit=limit,
        cursor=cursor,
    )

    return {
        "keys": keys,
        "next_cursor": next_cursor,
    }
````

**Comment by xqill275** [comment]
@mohammedFawzy0111 Hey, yeah, that would be great! I've been working on this for a little bit, and this is what I have

This is inside the cache class
```python
    def keys(self, limit: int, cursor: str | None = None) -> tuple[list[str], str | None]:
        with self.lock:
            if self.db is None:
                raise HTTPException(
                    status_code=500,
                    detail=f"Database file {self.filename} could not be opened and loaded",
                )
            all_keys = sorted(self.db.keys())

        if cursor:
            all_keys = [k for k in all_keys if k > cursor]

        page = all_keys[:limit]
        next_cursor = page[-1] if len(page) == limit else None

        return page, next_cursor
```

Then the endpoint looks like this:
```python
@app.get("/db/keys")
def get_keys(cache: Annotated[,limit: int = DEFAULT_KEYS_LIMIT,cursor: str | None = None, Cache, Depends(get_cache)]):
    if not 1 <= limit <= MAX_KEYS_LIMIT:
        raise HTTPException(
            status_code=400, detail=f"limit must be between 1 and {MAX_KEYS_LIMIT}"
        )

    keys, next_cursor = cache.keys(limit=limit, cursor=cursor)
    return {"keys": keys, "next_cursor": next_cursor}
```
I also added two variables inside the config file
```python
MAX_KEYS_LIMIT = 500
[...]

PR #148, opened 2026-09-15
added an end point to get all keys in the db

xqill275:
This is just an endpoint to get all keys in the DB.

The reason for this is that currently you can only access data if you already know the keys in the database

PR #156, opened 2026-09-15
Add GET /db/keys endpoint

mohammedFawzy0111:
Adds a paginated GET /db/keys endpoint (limit/offset, default 100, max 1000) to list stored keys.\nNo auth is added - SDB has none yet, so this stays consistent with the rest of the API rather than inventing a one-off security model.\nRefs #147


AnnoDomine:
Duplication of #148

Issue #160, opened 2026-09-15
Make database persistence crash-safe with atomic file writes


mohammedFawzy0111:
## Problem

"Cache.flush()" currently writes directly to the database file:

```python
with open(self.filename, "wb+") as f:
    f.write(json.dumps(self.db).encode())
```

If the process is interrupted while the file is being written, the existing ".sdb" file can be left truncated or otherwise corrupted.

"Cache.load()" already contains recovery logic for corrupted JSON files, which suggests that preserving the previous valid database during interrupted writes would be useful.

## Proposed solution

Change persistence to use an atomic write:

1. Serialize the database before modifying the existing file.
2. Write the serialized data to a temporary file in the same directory as the database.
3. Flush and "fsync()" the temporary file.
4. Atomically replace the existing database with "os.replace()".

The temporary file should be created in the same directory so that the replacement remains atomic.

The initial database creation in "_write_default()" should use the same mechanism.

## Expected behavior

If a write is interrupted or fails before the replacement occurs:

- the previous ".sdb" remains intact;
- the temporary file can be left behind or cleaned up;
- the next startup can still load the last valid database.

If the write completes successfully, the resulting ".sdb" contents should be unchanged from the current implementation.

MarkVDD:
This is mostly covered by PR #97, which adds WAL durability and atomic database compaction. The only remaining gap is initial database creation, which still writes directly to the final file.

PR #161, opened 2026-09-15
Make database persistence atomic

mohammedFawzy0111:
Makes database persistence crash-safe by writing database contents to a temporary file and atomically replacing the exisiting database file.

## Why
writing directly to the database file can leave the .sdb file truncated or corrupted if the process is interrupted during a write.

Atomic replacement ensures that the previous valid database remains intact until the new database has been completely written.

Refs #160


AnnoDomine:
See #97

AnnoDomine:
As the project is currently migrating to a python dependency, the file read and write logic will changed.
Additional this PR is far behind the current state i am closing it as `legacy`.

Issue #136, opened 2026-09-14
Wrong HTTP status code on successful PUT


aunrxg:
### Issue summary

This isn't an issue to be bothered with but good to acknowledge. When we create an entry in the db with PUT /db we are creating/uploading data in the db, so semantically we should use 201 CREATED (which represent a successfully created response) instead of currently used 200 OK (with denoted successful handling of the request).

### Steps to reproduce

1. Start the server.
2. Go to /docs (swaggerUI)
3. Try PUT with valid key and value

### Screenshot

<img width="1416" height="890" alt="Image" src="https://github.com/user-attachments/assets/6416f24f-f2da-4882-8c8e-58988e89f487" />

aunrxg:
It's a minuscule issue, i am not sure if i should raise a PR just for this. Somebody please fix it in their PR of a major issue thanks!

TynK-M:
To give more to this issue, [RFC9110](https://datatracker.ietf.org/doc/html/rfc9110#name-put):

> If the target resource does not have a current representation and the PUT successfully creates one, then the origin server MUST inform the user agent by sending a [201 (Created)](https://datatracker.ietf.org/doc/html/rfc9110#status.201) response. If the target resource does have a current representation and that representation is successfully modified in accordance with the state of the enclosed representation, then the origin server MUST send either a [200 (OK)](https://datatracker.ietf.org/doc/html/rfc9110#status.200) or a [204 (No Content)](https://datatracker.ietf.org/doc/html/rfc9110#status.204) response to indicate successful completion of the request. 

So the expected response should be 201 on first creation and 200 when modifying (204 would be if no content is returned, but we return the body).

PR #162, opened 2026-09-15
fix: corrent semantic HTTP status code for PUT #136

aunrxg:
## Description

This PR fixes the wrong status code issue highlighted in #136, if we are creating a new key-value pair in db the response status code should 201 CREATED instead of current 200 OK. The semantic difference between these two status code makes this PR relevant 

## Type of Change

- [x] Documentation update

## Key Changes

- added a simple is_new_key flag to check if the key is already present in cache.db
- based on this flag, update the response.status (fastapi)
- also updated the test cases to accept these changes
- added these explicitly in api.md

## Checklist

- [ ] Python changes pass `uv run ruff check .` and `uv run ruff format --check .`, or this PR does not change Python.
- [ ] Validation lists the checks or manual flows I ran and their results.


TynK-M:
I'm not an admin, but seems good to me.

src/seriousdb/main.py
```diff
@@ -29,8 +29,11 @@ def put(
     value: str,
     background_tasks: BackgroundTasks,
     cache: Annotated[Cache, Depends(get_cache)],
+    response: Response,
 ) -> str:
+    is_new_key = key not in cache.db
```
AnnoDomine: Thanks for tackling this! The semantic 201 vs 200 distinction makes total sense.
To keep `main.py` nice and clean, what do you think about borrowing a handy pattern from Django (`update_or_create`) and having `cache.insert()` return `(value, created)`?
This would make it more clear, thread-safe and removes a problem called "race condition".
What do you think?

AnnoDomine (changes requested):

src/seriousdb/main.py
aunrxg: Implemented this approach. cache.insert() now returns (value, created) while keeping the existence check and write under the lock to avoid the race condition. The PUT endpoint uses created to return 201 for new keys and 200 for updates. Thanks for the suggestion!

aunrxg:
hey @AnnoDomine due to multiple upstream changes the execution got a little messy, can you review the changes now

AnnoDomine (approved):

AnnoDomine:
@aunrxg 
Sorry, but we made a major change and added `ty` in the last merges.
Can you please update your branch and make alignments if they are neccessary?
If after this the CI is green, it can be merged.

aunrxg:
@AnnoDomine 
merged ty changes to branch, all test cases, type check and lint check passed locally

AnnoDomine (approved):

PR #163, opened 2026-09-15
Refactor request flow for improved clarity and readability

guenzel50-hue:
Made it look cleaner and more readable.


docs/architecture.md
```diff
@@ -12,5 +12,5 @@ The service starts with a default entry when `.sdb` does not exist. There is no
 
 1. FastAPI receives a request.
 2. The route loads the dictionary from `.sdb`.
-3. A `PUT` updates and rewrites the file; a `GET` reads the requested value; a `HEAD` only returns the header; a `DELETE` deletes the requested key.
+3. A `PUT` updates and rewrites the file.  A `GET` reads the requested value. A `DELETE` deletes the requested key.  A `HEAD` only returns the header.
```
AnnoDomine, line 15: This makes it more unreadable then befor and it does not reflect the current state.

AnnoDomine (changes requested):

docs/architecture.md
FriedAxons: @guenzel50-hue 

I checked the current `main` implementation against this change. The existing request-flow documentation is outdated in two places:

* `The route loads the dictionary from .sdb` is inaccurate because the database is loaded into the in-memory cache during application startup.
* `A PUT updates and rewrites the file` is also inaccurate because PUT/DELETE first modify the in-memory cache, then schedule `cache.flush()` as a FastAPI background task to persist the changes.

This PR only rearranges the existing sentence, so it doesn't address those discrepancies.

VihaanReddyM: I personally find the older version more readable as well. I don't think rearranging the sentence is necessary here.

More importantly, though, I agree that the request flow description needs to be updated to reflect the current implementation rather than just reworded. The cache is loaded during application startup and `PUT`/`DELETE` update the inmemory cache and schedule persistence through `cache.flush()`.

I think those implementation details should be corrected before we consider the documentation change complete.

PR #180, opened 2026-09-16
feat: Added count endpoint

### LINKED ISSUE #174: Add GET /db/count endpoint

There's currently no way to find out how many key-value pairs are stored without calling `GET /db/all` and counting the result client-side, this means transferring every stored value over just to get the size of the database.

### What outcome would help?

I'd like to propose a small and simple `GET /db/count` endpoint that returns the raw number of keys currently present in the database, following the same pattern as the existing `GET /db/all`. The response is a bare integer, consistent with how other endpoint return values are presented. I would like to open a PR for this issue, if the approach is reasonable

### Additional context

_No response_

VihaanReddyM:
## Description

This PR implements a new endpoint which return the total no of keys in the DB as an int. closes #174 

Updated documentation to reflect the same.

## Type of Change

- [x] Documentation update

## Key Changes

- Added new endpoint to return the count of all the keys
- Added tests for validation of the new endpoint
- Added documentation for the new endpoint

## Validation

```
uv run pytest tests
```

## Checklist

- [ -] I reviewed the diff and kept it focused on the described change.
- [ -] Python changes pass `uv run ruff check .` and `uv run ruff format --check .`, or this PR does not change Python.
- [ -] Changed behavior is covered by appropriate tests, including a regression test for bug fixes; any limitations are explained under Validation.
- [ -] Validation lists the checks or manual flows I ran and their results.
- [- ] Relevant documentation is updated, or no documentation change is needed.


tests/test_e2e.py
```diff
@@ -88,6 +88,15 @@ def test_delete_missing_key_returns_404(self):
         response = self.client.delete("/db", params={"key": "does-not-exist"})
         self.assertEqual(response.status_code, 404)
 
+    def test_count_returns_number_of_key_value_pairs(self):
```
AnnoDomine, line 91: Redudant tests.
The test in test_api.py fits and do the same incl. using pytest.
DRY => Don't repead yourself
Can you create a ~5000 item sized json and run the endpoint inside the terminal?

AnnoDomine (changes requested):

AnnoDomine:
Please update PR description. You additional made alignments to the documentation files.

tests/test_e2e.py
VihaanReddyM: Yea I did that because we had both the test files, I will remove it.

Also will run the 5000 test soon

VihaanReddyM:
```bash
va80@VA80s-MacBook-Air seriousdb % for i in $(seq 1 5000); do
  curl -s -X PUT "http://localhost:8000/db?key=key$i&value=value$i" > /dev/null
done
va80@VA80s-MacBook-Air seriousdb % curl -s "http://localhost:8000/db/count"
5001%                                                                                                                                                   va80@VA80s-MacBook-Air seriousdb % 
```

@AnnoDomine ^^^

tests/test_e2e.py
AnnoDomine: Nice. One small thing.
Add a test of a fresh database and remove the default key to get a count = 0
After this it is approved from my side :)

VihaanReddyM: ok I will remove the default key

AnnoDomine (approved):

PR #192, opened 2026-09-16
feat: ttl

notcr12dev:
## Description

This PR adds TTL support and the `DELETE /db` endpoint to manage key expiration and deletion.

Expired keys are cleaned up lazily when accessed, avoiding the need for a background thread. TTL metadata is stored separately from the actual values, keeping the API simple.

The implementation maintains backward compatibility with legacy `.sdb` files and only uses the new v2 format when TTL is enabled. Expired keys are also removed from disk on flush.

This provides a simple way to handle sessions, tokens, caching, basic rate limiting, and temporary locks.

## Type of Change

* [ ] Bug fix (non-breaking change which fixes an issue)
* [x] New feature (non-breaking change which adds functionality)
* [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
* [ ] Documentation update

## Key Changes

* Added TTL support for keys.
* Added lazy expiration of expired keys.
* Added `DELETE /db` endpoint for explicit key deletion.
* Kept backward compatibility with legacy `.sdb` files.
* Added separate TTL metadata storage.
* Removed expired keys from disk during `flush()`.
* No background thread is required for expiration.

## Validation

* Tested loading legacy `.sdb` files.
* Tested TTL expiration when accessing expired keys.
* Tested that expired keys are removed from `db` and `ttl`.
* Tested `DELETE /db` for explicit key deletion.
* Tested that expired keys are not written back to disk after `flush()`.
* Tested that data without TTL continues to use the legacy format.

## Checklist

* [x] I reviewed the diff and kept it focused on the described change.
* [x] Python changes pass `uv run ruff check .` and `uv run ruff format --check .`, or this PR does not change Python.
* [x] Changed behavior is covered by appropriate tests, including a regression test for bug fixes; any limitations are explained under Validation.
* [x] Validation lists the checks or manual flows I ran and their results.
* [x] Relevant documentation is updated, or no documentation change is needed.


AnnoDomine:
Please follow the contrubution guideline.
First open an issue, if the team approves the needing of this feature, then create a PR.
As pointed there:
> PRs that skip this process may be closed without review.

Closed as not follows the guideline

PR #210, opened 2026-09-17
Fix: remove unnecessary full-database copy in both /db/count and /db/bulk

### LINKED ISSUE #207: Prevent excessive memory usage in `/db/count` and `/db/bulk`
### What happened?

Both `/db/count` and `/db/bulk` create a shadow copy of the database before processing the request.

For `/db/count`, the entire database is copied only to calculate its length, even though `len(require_db(cache))` would be sufficient.

For `/db/bulk`, the entire database is copied even when the request asks for only a single key. The requested keys are then looked up in that copy.

As a result, these endpoints temporarily allocate memory proportional to the shadow copy of all the database for every request. With a sufficiently large database, this can cause significant memory spikes and may result in excessive memory usage or process termination due to an out-of-memory condition.

The behavior is observable even for requests that only access a small number of keys.

### Steps to reproduce

1. Start SeriousDB.
2. Populate the database with many records so that its in-memory representation is large (e.g. 2GB as I used).
    > Note: a large memory representation is not necessary, but it make it easier to observ the event.
3. Request the database count and observe the process memory usage while the request is handled. `/db/count` creates a full copy of the database before calculating its length.
4. Request a single key through the bulk endpoint, even though only one key is requested, `/db/bulk` first creates a full copy of the database and then performs the lookup on that copy.
5. Repeat either request while monitoring the process memory (for example with `top`, `htop`, or another process monitor). The temporary memory usage increases with the size of the database, rather than with the number of records actually requested.

The issue is deterministic and does not require concurrent requests.

### What did you expect?

I expect it to not allocate more memory than necessary and only do a copy when needed, e.g. for the `db/all`.

### Environment

seriousdb on commit: a799752eaa6592a7e194c6ebeed43655704d3d33
environment: WSL (Ubuntu distro) on Windows 10
python: uv 0.12.11 (x86_64-unknown-linux-gnu)

### Additional context

If possible I want to work on this issue.

rayvn-42:
Closes #207 

## Type of Change
- [ ] Documentation update

### Why
In both `/db/count` and `/db/bulk` `.copy()` is called on the full database before the request is processed, this allocate memory proportional to the size of the **entire** database no matter how small the request is. According to tests by @TynK-M in #207, this can cause real memory spikes depending on the size of the database even if only a single key is requested.

### What changed
- Removed `.copy()` in `count`. Since `len(require_db(cache))` is O(1) and doesn't need a copy.
- Removed `.copy()` in `get_bulk`. The dict comprehension already reads only the requested keys directly from the live dict, and builds a result sized to the request, not the whole database.
- Fixed `/db/bulk`'s `422` OpenAPI description, which incorectly described a "missing value", instead the `422` response in triggered if no key parameter is provided.
- Fixed a duplicate word in `/db/bulk`'s response description.
- Added type definition to `count`.

### Validation
- `uv run ruff format --check .`
- `uv run ruff check .`
- `uv run pytest tests`
- Reproducing the same scenario wouldn't trigger a memory spike

- [x] Python changes pass uv run ruff check . and uv run ruff format --check ., or this PR does not change Python.


src/seriousdb/main.py
```diff
@@ -179,9 +177,9 @@ def get_bulk(
         503: {"description": "The database file could not be opened and loaded."}
     },
 )
-def count(cache: Annotated[Cache, Depends(get_cache)]):
+def count(cache: Annotated[Cache, Depends(get_cache)]) -> int:
```
AnnoDomine, line 180: Return type removed?

AnnoDomine (changes requested):

src/seriousdb/main.py
rayvn-42: No, the previous count implementation had no return type definition, so I added. So that the code is more consistent.

AnnoDomine: My bad

AnnoDomine:
@rayvn-42 
I was wondered why the hole commits we already merged within another PR are listed again in this PR.
After a little research i found this: https://github.com/orgs/community/discussions/39031#discussioncomment-10199244
Can you please merge the main branch into this (even when it does not include any changes), to realign the commit history?
It is a mess in the history as then the changes is in the history twice.
In other cases it could even destroy the commit history.
For merge a PR into the main, we use the squash and merge option instead of "Create a merge commit" to not overload the commit history.
For now it is fine :)

PR #201, opened 2026-09-16
fix: allow empty PUT keys

wasiq345:
fixes: #196 
## Description

Allow empty string keys in `PUT /db` requests by removing the `min_length=1` validation. This fixes the incorrect rejection of valid PUT requests.

## Type of Change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update

## Key Changes

- Removed `min_length=1` validation from the PUT key parameter.
- Added regression coverage for empty PUT keys.
- Updated error-handling tests to expect successful empty-key PUT requests.
- No changes to GET, HEAD, or DELETE behavior.

## Validation

- `.`python` -m pytest -q`
- Result: `41 passed`

## Checklist

- [x] I reviewed the diff and kept it focused on the described change.
- [X] Python changes pass `uv run ruff check .` and `uv run ruff format --check .`, or this PR does not change Python.
- [x] Changed behavior is covered by appropriate tests, including a regression test for bug fixes; any limitations are explained under Validation.
- [x] Validation lists the checks or manual flows I ran and their results.
- [x] Relevant documentation is updated, or no documentation change is needed.

PR #211, opened 2026-09-17
Fixed: the empty string is now treated as a valid key in the DB API, consistent with a simple key-value store.

kamatealip:
## Description

This PR fixes the empty-key behavior for the database API by making the contract explicit and consistent across all endpoints. Empty-string keys are treated as valid keys in a key-value store, so repeated writes to `key=` overwrite the same value instead of creating duplicate entries.

The change also keeps the expected validation behavior intact:

- missing `key` parameters still return `422`
- non-empty keys that do not exist still return `404`
This addresses the inconsistency in the GET, HEAD, and DELETE handlers and adds regression coverage for the empty-key lifecycle.

## Type of Change

- [ ] Documentation update

## Key Changes
- Updated the `/db` key handling so empty-string keys are treated as valid keys
- Ensured repeated `PUT /db?key=&value=...` requests overwrite the same value rather than creating duplicates
- Made `GET /db?key=`, `HEAD /db?key=`, and `DELETE /db?key=`behave consistently with the same key contract
- Kept missing-key validation at `422` and non-empty missing-key lookups at `404`
- Added regression tests covering the empty-key lifecycle and overwrite behavior

## Validation

- Ran 
```bash
cd /home/msh/Documents/seriousdb && . .venv/bin/activate && pytest tests/test_api.py -q
```
- result  :- `20 passsed, 0 failed` with exit code `0`

## Checklist


AnnoDomine:
The issue is not linked to the PR.
This PR is a duplication of #201

Issue #213, opened 2026-09-17
Python API for seriousdb


danieldeer:
The HTTP-based API is not important any more. seriousdb will evolve into a serious storage layer.
Other Python projects shall be able to import seriousdb and call the get and set functions directly.
(and all the other functions which are currently only accessible via the HTTP interface).

VihaanReddyM:
How would we proceed with this? Shall the FastAPI endpoint be removed completely in the new implementation of the storage layer?
@danieldeer 

I personally thing if we are going to convert this project into a serious storage layer we should remove the HTTP endpoints.

rayvn-42:
Ok, just to clarify some things, I have a few questions:
- should the api use module functions like `import seriousdb as sdb; sdb.get(key)` or should it be object based like `db = sdb.Client(path)`, the second option can allow the user to handle multiple database instances at once?
- As for the `load()` which runs at startup and `flush()` which runs after each write, should the getter and setter functions auto load on initial use and flush synchronously, or the caller open/close with something like a context manager?
- should the http layer stay or get removed?
- Should other endpoints like `count` or `bulk` be added in initially or later on?

wasiq345:
I think moving to a Python API makes sense. Before we do that, it may be worth looking at Python’s built-in **shelve** module. It already provides a simple way to persist Python objects using key-value pairs, with an interface similar to a dictionary.

Comparing seriousdb with shelve would help us identify what additional features or benefits seriousdb should provide before we change its architecture.

VihaanReddyM:
@wasiq345 The goal of Daniel is to build a DB from shelve using python so I feel like shelve might not be the way to go since we want to make a DB engine from bottom up maybe using B-Trees.

danieldeer:
Yes, we somehow want to create our own database engine, probably also including serialization. So, we're not going to use the "shelve" module, or any other such high-level module.

danieldeer:
> Ok, just to clarify some things, I have a few questions:
> 
>     * should the api use module functions like `import seriousdb as sdb; sdb.get(key)` or should it be object based like `db = sdb.Client(path)`, the second option can allow the user to handle multiple database instances at once?
> 
>     * As for the `load()` which runs at startup and `flush()` which runs after each write, should the getter and setter functions auto load on initial use and flush synchronously, or the caller open/close with something like a context manager?
> 
>     * should the http layer stay or get removed?
> 
>     * Should other endpoints like `count` or `bulk` be added in initially or later on?


[...]

PR #227, opened 2026-09-18
Add Python module level API

mohammedFawzy0111:
## Description

Adds a synchronous Python API ( seriousdb.api ) so other projects can  import seriousdb  and call  get / set / delete / get_all / get_bulk / count / exists  directly — resolving #213. The module owns the shared  Cache ;  main.py  reuses it so HTTP and library access never diverge. Default DB file auto-loads on first use; writes flush synchronously.

## Type of Change

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update

## Key Changes

- Added a synchronous Python API for direct "seriousdb" usage without the HTTP server.
- Exposed "get", "set", "delete", "get_all", "get_bulk", "count", "exists", "load", "flush", and "is_loaded".
- Shared the same "Cache" instance between the Python API and HTTP layer.
- Added automatic loading of the default database file on first use.
- Made "set" and "delete" persist changes before returning.
- Added Python API tests.

## Validation

```text
uv run pytest
``` 
(all tests, incl. new  test_python_api.py ), 

## Checklist


src/seriousdb/__init__.py
```diff
@@ -1 +1,37 @@
-"""SeriousDB, a small key-value store served over HTTP with FastAPI."""
+"""SeriousDB, a small persistent key-value database.
+
+seriousdb can be used as a storage layer from other Python projects.
+>>> import seriousdb
```
fe-neu: I'm personally a big fan of ```import seriousdb as sdb```.

This keeps users from repeatedly having to type "seriousdb".

I know users can just do it themselves, but I feel communicating the intent is worth it.

src/seriousdb/api.py
```diff
@@ -0,0 +1,264 @@
...
+
+cache = Cache()
+
+_init_lock = Lock()
+
+
+def load(filename: str | Path = DB_FILE) -> None:
+    """Load the database from `filename`, replacing the current data.
+
+    If the file does not exist, it is created wit han empty database.
```
fe-neu: Typo here I think.

src/seriousdb/api.py
```diff
@@ -0,0 +1,264 @@
...
+    If it is not valid UTF-8 JSON or does not contain a JSON object, it is
+    renamed to ``<filename>.corrupt-<unix timestamp>`` and replaced wit an
+    empty database.
+
+    Parameters
+    ----------
+    filename : str or Path, optional Paht of the database file.
```
fe-neu: Typo here again

src/seriousdb/main.py
fe-neu, line 1: I'm not sure there is a clear vision on what should happen with the main.py file as of yet, so I'm curious what other people think.

In my opinion however it should either be gone all together or hold only a translation layer between FastAPI and our core logic (what you expose in api.py). Right now it's also holding logic and just reusing the Cache.

Open for a discussion on this though

fe-neu (commented):
Just my thoughts. Open for discussion though!

src/seriousdb/api.py
```diff
@@ -0,0 +1,264 @@
...
+
+from collections.abc import Iterable
+from pathlib import Path
+from threading import Lock
+
+from .cache import Cache, require_db
+from .config import DB_FILE
+
+__all__ = [
+    "cache",
```
AnnoDomine: Should not be accessable from outside the dependecy as it is an internal method.

src/seriousdb/api.py
```diff
@@ -0,0 +1,264 @@
...
+    "count",
+    "delete",
+    "exists",
+    "flush",
+    "get",
+    "get_all",
+    "get_bulk",
+    "is_loaded",
+    "load",
```
AnnoDomine: Should not be accessable from outside the dependecy as it is an internal method.

src/seriousdb/api.py
```diff
@@ -0,0 +1,264 @@
...
+__all__ = [
+    "cache",
+    "count",
+    "delete",
+    "exists",
+    "flush",
+    "get",
+    "get_all",
+    "get_bulk",
+    "is_loaded",
```
AnnoDomine: Should not be accessable from outside the dependecy as it is an internal method.

src/seriousdb/api.py
```diff
@@ -0,0 +1,264 @@
...
+
+from .cache import Cache, require_db
+from .config import DB_FILE
+
+__all__ = [
+    "cache",
+    "count",
+    "delete",
+    "exists",
+    "flush",
```
AnnoDomine: Should not be accessable from outside the dependecy as it is an internal method.

src/seriousdb/__init__.py
```diff
@@ -1 +1,37 @@
...
+    get,
+    get_all,
+    get_bulk,
+    is_loaded,
+    load,
+    set,
+)
+
+__all__ = [
+    "cache",
```
AnnoDomine: Should not be accessable from outside the dependecy as it is an internal method.

src/seriousdb/__init__.py
```diff
@@ -1 +1,37 @@
...
+    "count",
+    "delete",
+    "exists",
+    "flush",
```
AnnoDomine: Should not be accessable from outside the dependecy as it is an internal method.

src/seriousdb/__init__.py
```diff
@@ -1 +1,37 @@
...
+    "get",
+    "get_all",
+    "get_bulk",
+    "is_loaded",
```
AnnoDomine: Should not be accessable from outside the dependecy as it is an internal method.

src/seriousdb/__init__.py
```diff
@@ -1 +1,37 @@
...
+    "load",
```
AnnoDomine: Should not be accessable from outside the dependecy as it is an internal method.

AnnoDomine (changes requested):

AnnoDomine:
Please provide the output of pre-commit.

src/seriousdb/main.py
fe-neu: Okay Daniel just confirmed in #234

AnnoDomine: The removal of the HTTP API is a seperate part. We should not mix such a lot of changes to provide an overview.

AnnoDomine:
And please change the target branch to `development`

mohammedFawzy0111:
@AnnoDomine @fe-neu thank you for the feedback 
I tend to have typos when typing fast I will commit changes to fix them. 
I agree with @AnnoDomine for the internal methods 
But load is intended to be exposed as a way to load another .sdb file explicitly instead of default

src/seriousdb/__init__.py
mohammedFawzy0111: load is intended to be exposed as way to explicitly load non default sdb file

AnnoDomine:
> @AnnoDomine @fe-neu thank you for the feedback I tend to have typos when typing fast I will commit changes to fix them. I agree with @AnnoDomine for the internal methods But load is intended to be exposed as a way to load another .sdb file explicitly instead of default

Loading another db file is not in scope.
We only support default single db file in the first state.
Multi db files are pland for a future task, when the dependency have a stabe relesaed state.
In additional the hole cache system which is currently integrated will be completly refined and refactored.

mohammedFawzy0111:
@AnnoDomine done
And I also changed the base to development branch

mohammedFawzy0111:
Do you mean this by the pre-commit output

AnnoDomine:
> Do you mean this by the pre-commit output  

With `uv run pre-commit` it runs every required check in one :)
```bash
uv run pre-commit
      Built seriousdb @ file:///XXXXXXXXXXXXXXXXXXXXX/seriousdb
Uninstalled 1 package in 0.66ms
Installed 1 package in 1ms
uv-lock..............................................(no files to check)Skipped
ty.......................................................................Passed
pytest...................................................................Passed
ruff.....................................................................Passed
ruff-format..............................................................Passed
```

AnnoDomine (approved):

AnnoDomine:
Merged. Well done :)

PR #253, opened 2026-09-20
feat(api): Added prefix(scan) to Python API

### LINKED ISSUE #251: Add prefix-based key listing and scanning API

The current seriousdb API supports looking up individual keys, fetching a known set of keys with `get_bulk()`, or retrieving the entire database with `get_all()`.

There is currently no API for retrieving keys or values based on a key prefix.

For applications using structured key naming, such as:

```text
user:1001
user:1002
user:1003
session:abc
session:def
cache:homepage
```

a caller must either know the exact keys in advance or retrieve the entire database and filter it locally.

This becomes inconvenient as the number of stored keys grows and makes common namespace-based use cases unnecessarily expensive for the client.

A prefix-based key lookup/scan operation would allow users to retrieve only the entries belonging to a specific key namespace, such as all `user:*` keys, without requiring the entire database to be fetched first.

### What outcome would help?

A prefix-based scan operation would allow users to retrieve only the key-value pairs belonging to a specific key namespace without fetching the entire database.

For example:

```python
seriousdb.scan("user:")
```

could return:

```python
{
    "user:1001": "Alice",
    "user:1002": "Bob",
}
```

This would make it easier to work with structured keys such as `user:*`, `session:*`, or `cache:*` and avoid requiring clients to call `get_all()` and filter the results themselves.

The improvement should be available through the Python API and, where appropriate, the HTTP API, while preserving the existing behavior of the database.

It should also return a snapshot of the matching data so callers cannot accidentally modify the internal database state.

### Additional context

A current workaround is to use `get_all()` and filter the returned dictionary on the client side:

```python
data = seriousdb.get_all()

users = {
    key: value
    for key, value in data.items()
    if key.startswith("user:")
}
```

While this works, it requires retrieving the entire database even when the caller is only interested in a specific key namespace.

A dedicated prefix scan would make this use case part of the database API itself:

```python
users = seriousdb.scan("user:")
```

This could also be exposed through the HTTP API, for example:

```text
GET /db/scan?prefix=user:
```

The proposed feature is intentionally limited to prefix matching. Pagination, secondary indexes, regex queries, sorting, and changes to the persistence format can be considered separately.

Related work around database management and the Python API includes issue #220, which discusses the database/cache read and write architecture, and issue #213, which covers the Python API direction.

The implementation should remain compatible with the existing API and persistence model.

ShohinEshonov:
Closes #251

- [x] Documentation update

This pull request implements the feature requested in the issue #251. As for now, there's no way to fetch prefixed keys in a single request.

- Added scan function in `src/seriousdb/api.py'
- Added appropriate tests.
- Documentation for the function is written at `docs/api.md`

## Validation

- `uv run ruff format --check .`
- `uv run ruff check .`
- `uv run ty check .`
- `uv run pytest tests`

## Checklist


danieldeer:
PR was raised before Issue was accepted. Since the issue is closed, I also close this PR.

In the future, only raise PRs after being assigned implementation, otherwise you risk wasting your time.

Issue #249, opened 2026-09-19
Replace full-file rewrite persistence with a reliable write-ahead log


rayvn-42:
### What problem would this address?

`Cache` (as of this moment) holds the whole database in memory and rewrites the entire `.sdb` file on every write via `flush()`. That's the main cause for persistency and data loss issues (as stated in #160 and #90). Both of these issues are because of write-time corruption risk and non-atomic full-file rewrites. And in #97 attemepted a WAL-based fix but got halted because of missing tests for WAL, compaction and crash recovery.

### What outcome would help?

This issue proposes (and FYI, already has a working, fully tested implementation of the same general direction that #97 was going for). An append only write-ahead log `<file>.wal` that every write durably appends and fsyncs to immediately and with periodic compaction, appending the WAL file into the JSON file and clearing the WAL. `Cache`'s public methods (like `insert`, `select`, `delete`, `load`) keep their existing signatures, so that nothing that calls into `Cache`, needs to change.

A few scope decisions:
- Compaction trigger is a fixed write counter (defaults to 50), and it is not a size nor a time based trigger, open to suggestions if another approach is preffered.
- No cross-process locking, as for now the existing `cache.lock` only ever protects multiple threads withing a single process, multiple processes writing the same file concurrently is a seperate and much more difficult problem, which I don't attempt to solve within this issue. I have also updated `docs/persistence.md` to state this issue.
- `.sdb`'s on-disk format remains unchanged as a plain JSON snapshot, only addition is the WAL file.
- `flush()` becomes a no-op, and will be kept only because `main.py` (currently) still schedules it as a background task after each write. And durability now happens synchronously inso `insert`/`delete` instead, which is a strictly stronger guarantee than before when using `flush()`, since no background task is needed.

### Additional context

Using #222 as a guide to check in before any major changes take place, since this approach is already built and tested, I'm opening this issue for visibility and to discuss different approaches and methods. I'll be happy to modify and change my code if anything doesn't line up.

rayvn-42:
Not opening a PR until I'm told the approach is reasonable.

rayvn-42:
Also @AnnoDomine in #161 you said that persistence logic is changing, just wanted to know if this is the right approach.

izaaksarnecki1:
Hey! nice work, good call checkin in before creating a PR
[...]

PR #258, opened 2026-09-20
feat: replace full-file rewrite persistence with a write-ahead log

rayvn-42:
Relates to #97, #160, #90. Discussed in #249.

## Why
`Cache` rewrites the entire `.sdb` file on every write. That doesn't scale (see #254 for benchmarks). And its what causes the issues discussed in #90 and #160, although half of the issues, mainly atomic writes have already been resolved  (Thanks to @danieldeer for resolving the atomic write issues). This PR takes inspiration from #97 whilst adding the related tests and compaction which #97 wasn't able to achieve.

### `cache.py`:
- Every `COMPACTION_THRESHOLD` (defaults to 50) writes, the in-memory state is written to `.sdb` as a fresh snapshot and the WAL is cleared, both atomically (done by @danieldeer) and via temp-file-and-rename, in that order. So a crash at any point during compaction leaves either the old snapshot with the WAL intact, or the new snapshot with an empty WAL, when replaying and already-applied WAL entry is harmless since `set` and `delete` can be overlayed.
- `load()` replays anu WAL entries on top of the loaded snapshot, recovering any missed writes since last compaction. If a truncated/corrupt final entry (meaning a crash mid-append) is hit, the replay stops there rather than raising, every entry before that is recovered.
- `flush()` is now a no-op, only kept for backwards compatibility.
- `Cache`'s public method signatures remain unchanged.

### Tests (reworked `tests/test_cache.py`):
- Replay recovers writes made after last compaction.
- Compaction triggers at the threshold and leaves an empty WAL, and doesn't trigger before it.
- The compaction counter resets after each compaction.
- Replay is overlayable when a WAL entry was already applied last compaction to the snapshot before a simulated crash during compaction.

### Existing tests (updated due to moving to a per-write WAL append):
- `tests/test_python_api.py`: tests asserting persistence by reading `.sdb` immediately after one write now reload via a fresh `Cache()` instead, which correctly exercises WAL replay rather than assuming a removed synchronous full rewrite.
- `tests/test_logging.py`: `flush()`'s error-logging test replaced with a test that assert that `flush()` is a safe no-op.
- `benckmarks/test_persistence.py::test_flush`: now times `Cache.insert` (WAL append + fsync), the actual per-write cost, instead of timing a no-op.
- `benckmarks/test_benckmark_verification.py`: the fault-injection test retargeted from `cache.flush` to `cache._append_wal`, since that's the actual persisitence boundary.

### Docs:
- `docs/persistence.md` rewritten for the new (WAL) model, including correcting some existing inaccuracies (like it describing that `PUT`  as reloading the whole dict from disk on every request, which isn't true ever since `Cache` started living in memory.
- `cache.py`/`api.py` module docstrings updated from "flushed" to "updated in the background" to reflect the new synchronous WAL durablility.

## Validation
The new pre-commit validation already runs all the checks so no need to include them here

- [x] Documentation update

---


src/seriousdb/cache.py
```diff
@@ -70,6 +81,7 @@ def insert(self, key: str, value: str) -> tuple[str, bool]:
             db = require_db(self)
             is_new_key = key not in db
             db[key] = value
+            self._record_write({"op": "set", "key": key, "value": value})
```
MarkVDD, line 87: `insert()` updates the in-memory dictionary before `_append_wal()` succeeds; `delete()` has the same ordering. I simulated an append failure: the call raised, but the live cache contained the new value while a fresh cache did not. 

I think it is best if the state change gets applied only after the WAL write succeeds, with a regression test for append failure?

src/seriousdb/cache.py
```diff
@@ -175,30 +194,130 @@ def load(self, filename: str) -> None:
...
+                    continue
+                try:
+                    op = json.loads(line.decode())
+                except (json.JSONDecodeError, UnicodeDecodeError) as e:
+                    logger.warning(
+                        "Stopping WAL replay at truncated/corrupt entry in %s (%s)",
+                        self.wal_filename,
+                        e,
+                    )
+                    break
```
MarkVDD: Replay stops at a torn WAL entry but leaves that partial entry in place. I reproduced this by truncating the last record, loading the database, and successfully inserting another key. The new record was appended to the partial line; on the next load, replay stopped there again and the successful insert disappeared. 

Could recovery truncate the WAL to the last valid record before accepting new writes, or refuse writes until it is repaired?

src/seriousdb/cache.py
```diff
@@ -175,30 +194,130 @@ def load(self, filename: str) -> None:
...
+
+        Parameters
+        ----------
+        op : dict
+            A JSON-serializable write operation.
+        """
+        self._append_wal(op)
+        self._writes_since_compact += 1
+        if self._writes_since_compact >= COMPACTION_THRESHOLD:
+            self._compact()
```
MarkVDD: It looks like `insert()` can raise an error even after the write is durable, since compaction runs post-fsync.

Should we isolate compaction failures from the primary write result so callers get an accurate signal on durability?

src/seriousdb/config.py
```diff
@@ -20,4 +23,5 @@
 load_dotenv()
 
 DB_FILE = os.getenv("SERIOUSDB_DB_FILE", ".sdb")
+WAL_FILE = os.getenv("SERIOUSDB_WAL_FILE", ".sdb.wal")
```
MarkVDD: `SERIOUSDB_WAL_FILE` is exposed as configuration, but `Cache.load()` always uses `f"{filename}.wal"` and never reads this setting. 

Could we wire it through or remove the unused option? As written, setting it has no effect.

docs/persistence.md
```diff
@@ -1,25 +1,20 @@
...
-By default, data is stored in `.sdb` in the process working directory. Set `SERIOUSDB_DB_FILE`
-before importing the package to use a different path; see [configuration](configuration.md).
+Data is stored in a local file named `.sdb` in the process working directory, alongside a write-ahead log file named `.sdb.wal`
 
-The file contains a JSON object. Loading a missing file creates an empty database (`{}`). Reads use
-the in-memory dictionary without reloading changes from disk.
+The `.sdb` file contains a serialized Python dictionary written with the standard-library `json` module. On first startup, the application creates it with:
 
 The [architecture guide](architecture.md#call-flow) shows when the API loads and writes its cache.
 Direct `Cache` mutations require an explicit `flush()`.
```
MarkVDD: This still says direct Cache mutations require `flush()`, but this PR makes `flush()` a no-op and persists changes in `insert()` and `delete()`. Could this section be updated? The sentence above it also appears unfinished.

src/seriousdb/cache.py
rayvn-42: I understand. I think perhaps the safest option is to remove torn and uncomplete entries completely, although this can result in data loss, I'm not sure if there's a better approach to this.

src/seriousdb/config.py
rayvn-42: I think removing it is a better approach since, the sdb and wal files always need a similar basename ("basename.sdb" and "basename.sdb.wal") and exposing just adds a layer of complexity for the user.

docs/persistence.md
rayvn-42: I will fix that I just forgot to update it.

src/seriousdb/config.py
Airbag65: To add onto this, I think we should add `*.sdb.wal` to the .gitignore file. `*.sdb` is already ignored so it would make sense, and we don't want someones half-baked test data to accidentally make it into the repo

src/seriousdb/cache.py
Airbag65: I noticed this as well

src/seriousdb/config.py
rayvn-42: I'll also consider adding that.

src/seriousdb/cache.py
```diff
@@ -175,30 +194,130 @@ def load(self, filename: str) -> None:
...
+        Parameters
+        ----------
+        op : dict
+            A decoded WAL entry, as produced by :meth:`_append_wal`.
+        """
+        db = require_db(self)
+        if op.get("op") == "set":
+            db[op["key"]] = op["value"]
+        elif op.get("op") == "delete":
+            db.pop(op["key"], None)
```
Airbag65, line 275: Might not be a necessary modifications (get-ops on dicts are fast enough as they are) but I would have made the `op` into it's own variable and then do comparisons on that. Something like this:
```python
db_op = op.get("op")
if op == "set":
    ...
elif op == "delete":
    ...
```
Or perhaps even with a match statement to add explicit exhaustiveness:
```python
db_op = op.get("op")
match db_op:
    case "set":
        ...
    case "delete":
        ...
    case _:
        # unrecognized OP
```

This is just my way of seeing thing things however, I might be overthinking a very simple thing

src/seriousdb/cache.py
rayvn-42: I've been looking into this and the standard parctice is to just truncate the corrupt entry.

Airbag65:
Overall I'd say it looks like a very solid WAL system. Some small things to look over (see above, @MarkVDD pointed out some good things), but overall I like your implementation. Good work!

rayvn-42:
@Airbag65 Thanks. I am looking into how to fix the issue pointed out by @MarkVDD.

src/seriousdb/cache.py
rayvn-42: I think looking up the operation only once is better and more efficient than my implementation, I will add it.

rayvn-42:
So I fixed all the suggested changes, added regression tests in `tests/test_cache.py`. And fixed all docmentation mistakes.

src/seriousdb/api.py
```diff
@@ -3,6 +3,14 @@
...
 import the package and call the functions exported here directly.
 
+All functions operate on a single, module-level
+:class:`~seriousdb.cache.Cache` that is shared with the HTTP server. The
+database file (by default ``.sdb``, see :mod:`seriousdb.config`) is loaded
+automatically on the first call; use :func:`load` to load a different file
+explicitly.
+
+:func:`set` and :func:`delete` durably append to the write-ahead log before
+they return, so a successful call is persisted even before the next compaction.
```
danieldeer: Please remove this entire block of documentation. Users don't need to know how the DB internally works (e.g. through Cache), and also the documentation is outdated (there is no HTTP server any more).

It is fine to let users know about the guarantees of the API though ("A successful call is persisted")

danieldeer:
While the changes are clear and simple I would really like to not mix the WAL with the cache, as they are (from an object oriented programming perspective) quite different things.

I think we should create a new class `wal.py` that contains a write-ahead log class, and all associated functions. 

What do you think @rayvn-42 ?

src/seriousdb/cache.py
```diff
@@ -176,30 +222,168 @@ def load(self, filename: str) -> None:
...
+        """No-op, kept for backward compatibility.
+
+        Durability is now handled per-write via the write-ahead log (see :meth:`_append_wal`),
+        so nothing needs to happen here.
+        """
+        return
 
-        The file is overwritten with the full database. Does nothing if no
-        database has been loaded.
+    # ------ Write-ahead log internals -----------------------------------------#
```
danieldeer: These section comments usually become outdated after some time (because people will append new methods that are unrelated to the WAL). Also dedicating a full section to new functionality (WAL) indicates that the functionality should be refactored into a dedicated class (e.g. `wal.py`)

rayvn-42:
I think that's a great idea. And it would even make detecting bugs much easier. I will implement this.

rayvn-42:
WAL implementation seperation from cache file is almost done. I'm currently adding and seperating the wal tests into `tests/test_wal.py`.

rayvn-42:
So the WAL is split out from `Cache` into its own `wal.py`. Thanks to @danieldeer for suggestion.

Added: `wal.py` with a standalone `WriteAheadLog` class (`append`/`replay`/`clear`), and `test_wal.py` which directly tests `wal.py` without involving the `Cache`.

Changed: `Cache` now holds a `WriteAheadLog` instance instead of managing the log file itself. Updated `test_cache.py` to execlude `wal` tests, and the new `cache.wal.filename` and `cache.wal.append()`. Also fixed a couple docstrings/comments.

I think these updates should cover what was talked about.

src/seriousdb/api.py
rayvn-42: I think the new commit resolves this completely.

src/seriousdb/cache.py
rayvn-42: So, basically cache still needs orchestrating functions to talk efficiently with WAL (basically functions that are best kept within the `cache.py` file, all other WAL related functions have been moved to their dedicated `wal.py` file.

danieldeer:
@rayvn-42 is the PR ready to review? It is still marked as Draft.

rayvn-42:
Just resolving some merge conflicts from main since it has been updated, and then I'll make it ready for review.

rayvn-42:
@danieldeer its done. All fixed.

VihaanReddyM (approved):
Nice implementation!

src/seriousdb/cache.py
```diff
@@ -176,30 +207,102 @@ def load(self, filename: str) -> None:
...
+
+        Parameters
+        ----------
+        op : dict
+            A decoded write-ahead entry entry, as produced by :meth:`~seriousdb.wal.WriteAheadLog.append`.
+        """
+        db = require_db(self)
+        db_op = op.get("op")
+        if db_op == "set":
+            db[op["key"]] = op["value"]
```
danieldeer: I find the acting on string fields of the dict rather unsafe. How about we make a WAL-Entry a class, rather than a dict with string fields?

Then we could do:
```
db: dict[str, str] = require_db(self)
key: str = op.get_key()
value: str = op.get_value()
db[key] = value
```

And everything would be perfectly protected by the type checker.

Currently, if I am not mistaken, people have to search the implementation to understand how to construct the fields of a WAL entry.

src/seriousdb/wal.py
```diff
@@ -0,0 +1,113 @@
...
+
+        Raises
+        ------
+        OSError
+            If the temporary or final files cannot be written.
+        """
+        dir_name = os.path.dirname(self.filename) or "."
+        with tempfile.NamedTemporaryFile("wb", dir=dir_name, delete=False) as tmp_file:
+            pass
+        os.replace(tmp_file.name, self.filename)
```
FriedAxons: Could the temporary file be cleaned up if `os.replace()` fails?

Both `WriteAheadLog.clear()` and `Cache._compact()` create temporary files with `delete=False` and then replace the destination, but neither appears to clean up the temporary path if `os.replace()` raises. I tested both failure paths by forcing `os.replace()` to raise, and in each case the temporary file remained in the directory.

Could we add cleanup for the temporary file on the replacement failure path?

FriedAxons (commented):
One documentation issue I noticed while reviewing the persistence changes:

`docs/architecture.md` currently says that `set` and `delete` immediately flush the in-memory dictionary back to `.sdb`. With the persistence model introduced by this PR, writes are instead appended and fsynced through `.sdb.wal`, while the `.sdb` snapshot is updated during compaction.

Could this section be updated or removed as part of the WAL change so the architecture documentation does not describe the old persistence model after this PR is merged?

rayvn-42:
I will rewrite it so that reflects the new changes. Thank you for pointing it out.

tests/test_wal.py
```diff
@@ -0,0 +1,102 @@
...
+
+
+def test_append_writes_one_json_line_per_entry(wal, wal_path):
+    wal.append({"op": "set", "key": "a", "value": "1"})
+
+    with open(wal_path, "rb") as f:
+        content = f.read()
+
+    assert (
+        content == json.dumps({"op": "set", "key": "a", "value": "1"}).encode() + b"\n"
```
MarkVDD: Startup recovery handles a torn final record now. What happens if an append partially writes and raises while the process keeps running? It looks like the next write could append directly after that incomplete record. Could we cover that case with a test?

.pre-commit-config.yaml
```diff
@@ -1,3 +1,6 @@
+default_language_version:
+  python: python3.13
```
MarkVDD: Could we drop the Python 3.13 pin? The development guide supports Python 3.11+, but this requires contributors to have Python 3.13 installed to run Python-based hooks.

.gitignore
```diff
@@ -220,3 +220,4 @@ __marimo__/
 
 # Database
 .sdb
+.sdb.wal
```
MarkVDD: I’d prefer `*.sdb.wal` so custom database names such as `test.sdb.wal` are ignored too.

rayvn-42: Should it be done for `.sdb` too?

rayvn-42:
All reviewed issues are fixed.

.gitignore
MarkVDD: I'd like that 👍🏽

rayvn-42: I'll add it.

rayvn-42:
Fixed new benchmark processes implementation to work with `WAL` architecture.

tests/test_wal.py
FriedAxons: I tested this against the current PR head (`75c6dfc`) by forcing an `append()` to write partial bytes and then raise while keeping the same `WriteAheadLog` instance alive.

The next `append()` repaired the torn tail back to the last known-good offset before writing the new record. Final replay returned the original record plus the later successful write, so the failed partial record was not replayed.

So the current implementation appears to handle the long-running-process case correctly. I don't see a regression test that exercises an actual partial `append()` followed by an exception, though, so I agree it would be useful to cover that specific failure path in `tests/test_wal.py`.

src/seriousdb/cache.py
```diff
@@ -176,30 +207,106 @@ def load(self, filename: str) -> None:
                     )
                     self.db = _write_default(filename)
             self.filename = filename
+            self.wal = WriteAheadLog(f"{filename}.wal")
+            self._writes_since_compact = 0
```
FriedAxons: Could we clarify how the compaction threshold is intended to behave across a restart?

`load()` resets `_writes_since_compact` to `0` before replaying the remaining WAL entries, so those replayed writes are not reflected in the counter. I reproduced 49 WAL records, reloaded the database, and then performed one additional write. At that point the WAL contained 50 records, but compaction did not occur and the snapshot remained `{}`.

The PR describes compaction as occurring every `COMPACTION_THRESHOLD` (50) writes, while startup replay recovers writes made since the last compaction.

Are replayed WAL writes intended to count toward the compaction threshold across a restart? I'm not assuming a particular implementation.

rayvn-42: So, yeah that is a bug and not intended behavior. `load()` resets the counter to 0 instead of the length of the WAL entries. In your case, when the database reloaded, it set `_writes_since_compaction` to `0` so performing an addtional write only brought the counter to `1`. Should be a simple fix, just updating the `_writes_since_compaction` to `len()` of the current WAL entries.

rayvn-42: So I have fixed the issue, and implemented a test in `test_cache.py` to recreate the scenario.

src/seriousdb/cache.py
rayvn-42: I'll add a seperate class for WAL entry.

danieldeer:
Thanks a lot for all the work done so far. To get this feature in I would propose to accept it today, as soon as the WAL Entry class is added.
Any further issues with the WAL can be raised separately.

Would you be interested in being a collaborator on this project @rayvn-42 ?

rayvn-42:
> Thanks a lot for all the work done so far. To get this feature in I would propose to accept it today, as soon as the WAL Entry class is added. Any further issues with the WAL can be raised separately.
> 
> Would you be interested in being a collaborator on this project @rayvn-42 ?

I would love to, this project has been one of the few projects that genuinely intrigued me. Thank you @danieldeer and the team for being so helpful and responsive.

src/seriousdb/cache.py
rayvn-42: I have added type WAL entries for setting (`SetEntry`) and deleting (`DeleteEntry`). All tests and docstrings have been appropriately updated.

PR #260, opened 2026-09-20
fix(cache): preserve corrupt backups on timestamp collision

### LINKED ISSUE #255: Corruption recovery can overwrite an existing backup within the same second
## What happened?

When `Cache.load()` recovers a corrupt database file, the backup path is generated using:

```python
backup = f"{filename}.corrupt-{int(time.time())}"
```

Because the filename only has second-level timestamp precision, two corruption recoveries within the same second can generate the same backup path.

The later recovery then overwrites the earlier `.corrupt-*` backup, causing the previous corrupt database contents to be lost.

## Steps to reproduce

Run this from the repository root:

```bash
PYTHONPATH=src uv run --no-project --with python-dotenv python - <<'PY'
from pathlib import Path
from tempfile import TemporaryDirectory
from unittest.mock import patch

from seriousdb import api

with TemporaryDirectory() as raw_dir:
    directory = Path(raw_dir)
    database = directory / "database.sdb"
    payloads = (b"FIRST_CORRUPT_PAYLOAD", b"SECOND_CORRUPT_PAYLOAD")

    # Make both recoveries use the same second for backup naming.
    with patch("seriousdb.cache.time.time", return_value=1_700_000_000):
        for number, payload in enumerate(payloads, start=1):
            database.write_bytes(payload)
            api.load(database)

            backups = {
                path.name: path.read_bytes()
                for path in directory.glob("database.sdb.corrupt-*")
            }
            print(f"After recovery {number}: {backups}")

    assert all(payload in backups.values() for payload in payloads), (
        "An earlier corruption backup was lost"
    )
PY
```

## What did you expect?

Each corruption recovery should preserve the corrupt database it is recovering.

A later recovery should not overwrite an existing `.corrupt-*` backup.

The exact naming strategy can remain an implementation detail, but backup paths should be collision-safe, for example by using a UUID, a nanosecond-resolution timestamp, or another uniqueness mechanism.

## Environment

*No response*

## Additional context

The script intentionally corrupts the same database file twice and freezes the timestamp to reproduce a same-second backup-name collision deterministically.

It demonstrates backup loss during recovery, not the cause of the original corruption.

Dexr1te:
## Description

Fixes #255 .

Prevent corrupt database backups from being overwritten when multiple recovery attempts occur within the same Unix second.

## Type of Change

- [x] Documentation update

## Key Changes

- Select the first unused `.corrupt-<timestamp>` backup path.
- Append numeric suffixes such as `-1`, `-2`, etc. when collisions occur.
- Preserve existing corruption backups instead of overwriting them.
- Add deterministic regression tests for repeated same-second recovery.
- Update documentation to describe the backup naming behavior.

## Validation

- `uv run pytest` — 65 passed
- `uv run ruff check .` — passed
- `uv run ruff format --check .` — passed
- `uv run ty check .` — passed
- `uv run pre-commit run --all-files` — all hooks passed

## Checklist


danieldeer:
All clear, thank you. Merging this PR!

PR #263, opened 2026-09-20
test: add multi-process benchmarks

### LINKED ISSUE #257: Add multi-process performance benchmarks

The current benchmarks cover reads, loading, flushing, and shared threads, but not multiple OS processes accessing the same database.

It would be valuable to add these benchmarks before the WAL and read/write coordination implementations land, so we have a baseline for measuring their impact.

### What outcome would help?

Add read-only multi-process benchmarks that measure:

- Cold database loading
- Steady-state reads
- Increasing process counts
- Correctness of returned data

Concurrent writes are out of scope until the write and concurrency semantics are defined.

Related: #214, #220, #249

### Additional context

_No response_

MarkVDD:
## Description

Adds multi-process benchmarks to establish a baseline for future concurrency improvements. Closes #257.

## Type of Change

- [x] Documentation update

## Key Changes

 - Benchmark cold loads and steady reads across 1, 2, 4 and 8 processes, checking results each round.
  - Add opt-in concurrent writes that record failures such as lost updates and Windows replacement conflicts.
  - Consolidate benchmark usage and implementation details into one guide.

  Example:
  ```bash
  uv run --locked --group benchmark -m benchmarks -k process_reads
  ```

## Validation

Run from the repository root with `uv`. Validated on Windows:

  - `uv run pytest` — 132 passed.
  - `uv run --locked --group benchmark pytest benchmarks --benchmark-disable` — 60 passed, 12 opt-in cases skipped.
  - `uv run --locked --group benchmark -m benchmarks -k "process_reads or process_writes" --multiprocess-writes -rx` - 27 passed, 9 expected write failures.
  - `uv run ruff check .`, `uv run ruff format --check .`, `uv run ty check .` passed.

Linux and macOS were not tested locally.

## Checklist


benchmarks/README.md
VihaanReddyM: I think this should be added to the DOCS or if the documentation has already been added to the docs then this just feels redundant. 

@danieldeer what do you think?

danieldeer:
Is it technically necessary for the benchmarks to use the Cache directly, rather than going through the API? We're, after all, testing process performance here, and processes also access `seriousdb` via the API.
Going directly into the cache could cause complication later in case we refactor the layers below the cache.

I also find the README a big addition. And the occasional em-dash (—) in the PR indicates AI usage. How carefully was this PR checked by a human before submission?

I am not inclined to review >600 lines of AI generated code.

MarkVDD:
> Is it technically necessary for the benchmarks to use the Cache directly, rather than going through the API? We're, after all, testing process performance here, and processes also access `seriousdb` via the API. Going directly into the cache could cause complication later in case we refactor the layers below the cache.
> 
> I also find the README a big addition. And the occasional em-dash (—) in the PR indicates AI usage. How carefully was this PR checked by a human before submission?
> 
> I am not inclined to review >600 lines of AI generated code.

AI helped during implementation whenever I had questions, the vast majority of the code was written by myself, and I reviewed all of it before opening the PR. The automated tests and checks also passed.

The em dashes are in the PR description, which I did generate with AI, but checked to make sure it was complete.

Using Cache directly wasn’t technically necessary. I followed the existing benchmarks, which also use it directly. Your concern about coupling these benchmarks to the internal implementation makes sense, and I’ll switch them to the API.

For the README, would you prefer a shorter guide focused on the new benchmarks? I think documenting the full suite adds value, but I’m happy to discuss the scope if this is too large.

MarkVDD:
Running the following command:
```bash
uv run --locked --group benchmark -m benchmarks -k "process_reads and 1000x32B"
```
Results in these benchmarks:

image

API benchmark results on Windows. Cold runs reload through `api.load()`; resident runs reuse loaded data. Each round performs 1,000 total lookups across the selected processes. Timings include process communication but exclude startup.

Benchmark command for the writes:
```bash
uv run --locked --group benchmark -m benchmarks -k "process_writes and 1000x32B" --multiprocess-writes -rx
```

Results in this:
image

API write benchmark on Windows, using a database of 1,000 entries with 32-byte values. Each round attempts 100 updates through `api.set()`, which persists each write.

The single-process case passed. Runs with 2, 4, and 8 processes detected Windows replacement conflicts and lost or incorrect persisted updates, reported as expected failures (XFAIL).

Timings are retained as a baseline, but the shorter concurrent timings do not represent successful write throughput because those runs failed correctness checks.

benchmarks/README.md
```diff
@@ -0,0 +1,179 @@
+# Benchmarks
```
danieldeer, line 1: This README.md is doomed to get outdated as soon as benchmark tests are added or updated.

Please, therefore, reduce the size of the README. Keep it focused to general information about the benchmarks, and user instructions to run the benchmarks.

The description of the individual tests should be moved into the actual test functions as docstrings.

Once this is resolved, set it to ready, ping me, and I will approve and merge this PR.

MarkVDD:
@danieldeer Ready for another review! I've shortened the README and moved the benchmark descriptions into the test functions' docstrings.

danieldeer (approved):
Fine!

PR #264, opened 2026-09-20
feat: Add pagination on get_all()

Asbjoern09:
## Description

It adds optional pagination to the get_all() function, making it `def get_all(page: int | None = None, size: int | None = None)`. 

It means that when calling the function, you can split the get_all into chunks (AKA pages), specifying the page and size of each page.

I've had many cases when using APIs and databases, where i'm missing pagination, so i thought it would make sense to add on a get_all, which can potentially be very heavy function call for a larger database.

## Type of Change

- [ ] Documentation update

## Key Changes

- Adds optional pagination to get_all(), nothing breaking, as it defaults to previous handling if no pagination is supplied in function call

## Validation

- Executed a Python test script populating 30 database keys to verify multi-page scenarios.
- Tested page splitting and dynamic page sizes across sequential pages.
- Confirmed default `get_all()` behavior remains backward compatible when parameters are omitted.
- Added a unit test to verify functionality

## Checklist


src/seriousdb/api.py
```diff
@@ -172,9 +173,11 @@ def exists(key: str) -> bool:
         return key in require_db(cache)
 
 
-def get_all() -> dict[str, str]:
+def get_all(page: int | None = None, size: int | None = None) -> dict[str, str]:
     """Return a snapshot of every key-value pair in the database.
 
+    get_all supports optional pagination with page:int and size: int
```
mohammedFawzy0111, line 179: Docstring doesn't follow the project's NumPy convention.  page  and  size  aren't documented in a  Parameters  section

```text
Parameters
----------
page : int, optional
    Page number, starting at 1. ...
size : int, optional
    Number of entries per page. ...
```

src/seriousdb/api.py
```diff
@@ -188,9 +191,20 @@ def get_all() -> dict[str, str]:
...
     _ensure_loaded()
+    
     with cache.lock:
-        return require_db(cache).copy()
+        db = require_db(cache)
+        
+        if page is None or size is None:
+            return db.copy()
+
+        page = max(1, page)
```
mohammedFawzy0111, line 201: Edge semantics are silently clamped.  page = max(1, page)  means  
```python
get_all(page=0, ...)
```
  or a negative page quietly returns page 1 instead of surfacing a caller bug. Clamping is a defensible choice, but it should be a deliberate, documented decision (alternatively raise  ValueError ).

danieldeer:
No issue was raised before opening this PR, we won't merge it like this. Please check contribution docs before opening PRs

